PR #20, opened 2015-11-25
Test that asserts that the reposition of the dropdown is runloopaware

cibernox:
@snewcomer I copied your test but this seems to work both with and without the runloop aware change.

It's not bad to have it but seems that it's only asserting that the MutationObserver is being called properly. Anyway, there is no such thing as too much tests, so i'm merging this.


snewcomer:
Yep.  We were trying to fail the test, i.e. by replicating the problem in our app that repositions the dropdown and gives us a run loop error.  Whether or not `Ember.run(this, this._runloopAwareRepositionDropdown);` is wrapped in a run loop or not, it still passes.  Anyways, thanks for merging it in!  Maybe solve it at a different time.

PR #22, opened 2015-12-02
[DEPENDENCY]: using getOwner to support ember 2.3

toranb:
I've been fixing a few of our addons and this was another getOwner update :)

If you pull this in, when do you think a beta release of power-select could ship (to include this) ?


cibernox:
I can bump a minor version tomorrow. 

Thanks

toranb:
@cibernox thank you much !!!

PR #24, opened 2015-12-06
Allowing to open close component passing `opened=true/false`

cibernox:
This finally allows to pass the component a property to remotely control the component. 

When the property passes from falsey to truthy the component opens, and when goes to falsey it closes.

It is absolutely equivalent to open/close the component manually (the onOpen/onClose actions are invoked if provided), but at least for render the component already opened it avoids a repaint (render it closed and then re-render it opened)


miguelcobain:
@cibernox, you're the man.

PR #36, opened 2016-01-07
when rendered in place don't handle root mouse click

chrisgame:
Currently if renderInPlace is set to true no content is rendered into the wormhole so when the handler of route mouse down is fired it errors because .ember-basic-dropdown-content is not anywhere on the page.

### DIFF
--- a/addon/components/basic-dropdown.js
+++ b/addon/components/basic-dropdown.js
@@ -133,8 +133,10 @@ export default Component.extend({
   },
 
   handleRootMouseDown(e) {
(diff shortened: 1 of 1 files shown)


cibernox:
In theory, `ember-basic-dropdown-content` should be on the page too if the component is rendered in place. And clicking outside the component should close it.

Any reproducible example? I think the problem must be elsewhere.

chrisgame:
fixed by https://github.com/cibernox/ember-basic-dropdown/commit/9fe33b22da5ec460838e1ad85441d649279be807

thanks @cibernox

PR #40, opened 2016-01-23
Added labelledby, describedby and required aria properties to the trigger

martndemus:
See cibernox/ember-power-select#254


cibernox:
Question: `{{basic-dropdown ariaLabelledBy="foobar"}}` or `{{basic-dropdown aria-labelledBy="foobar"}}`.

The second seems closer to the html spec.

martndemus:
I think `ariaLabelledBy`, but I don't really have a good reason or anything for it, it just feels more natural.

cibernox:
I, in general, prefer to name things in my components as close as possible to the equivalent html spec, however in this case i've seen that ember has out of the box support for `aria-role` via the `ariaRole` property, so let's be consistent with ember here.

cibernox:
Published in 0.8.3

PR #41, opened 2016-01-24
Add support for `role`, `aria-required` and `aria-invalid`

martndemus:


cibernox:
Do you need a release?

martndemus:
Nah, I'll work with npm link until im done adding aria stuff

PR #47, opened 2016-02-10
Implement all role & aria properties in RFC

cibernox:
See https://github.com/cibernox/ember-power-select/issues/293#issuecomment-181419090

 - [x] `role=button`
 - [x] `aria-haspopup=true` tells AT to treat the trigger as a host for conditional content
 - [x] `aria-expanded=true/false` tells AT that conditional content is expanded
 - [x] `aria-pressed=true/false` tells AT that the `button` is in a pressed or 'true' state (technically different from whether the conditional content is expanded, but in practice usually the same).
 - [x] `aria-controls=<id of dropdown>` is all we need since there is no requirement that the trigger element own or have a direct DOM relationship to the dropdown

### DIFF
--- a/addon/components/basic-dropdown.js
+++ b/addon/components/basic-dropdown.js
@@ -9,6 +9,7 @@ export default Component.extend({
   layout: layout,
   disabled: false,
   renderInPlace: false,
+  role: 'button',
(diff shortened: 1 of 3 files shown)


jmacqueen:
Woo!

PR #61, opened 2016-03-14
Basic support for transitions in the dropdown

cibernox:
When the dropdown is opened, it has a class `ember-basic-dropdown--transitioned-in` class that is removed after the component has been rendered.

This allows this simple css snipped to work:

```scss
.ember-basic-dropdown-content.fade-in-dropdown {
  transition: opacity .2s;
  opacity: 0;
  &.ember-basic-dropdown--transitioned-in {
    opacity: 1;
  }
}
```

The opposite animation is a bit more complex, since you can't animate the removal of DOM from the document. The approach is slightly different in this case. First the element gets a `ember-basic-dropdown--transitioning-out` class. Then, in the `afterRender` (once that class has been applied to the DOM) if the dropdown has any transition happening (this is done by checking if the `transition-duration` is longer than `0s`) this adds a listener using the `transitionend` event listener.

Once that event is fired the dropdown is closed for good.

If there is no animations the component is closed right away.

This means that when a dropdown has a transition, it might be opened simultaneously with another one, whereas before this PR, dropdowns were mutually exclusive (they still are if we ignore the transitions)


Dremora:
Have you considered using [animationend](https://developer.mozilla.org/en-US/docs/Web/Events/animationend) event to determine when animation ends instead of relying on an additional attribute?

cibernox:
@Dremora I haven't, and that sounds good. Does it work with transitions apart from animations?

Dremora:
AFAIK ng-animate is using it. Haven't worked with ember-fire, but I presume it works in a similar way.

cibernox:
@Dremora I used the `transitionend` event. The tricky part was to know when to use it and when just close the dropdown right away, because for that I need to know if there is any transition beforehand, but I found a way.

PR #62, opened 2016-03-15
Better animations

cibernox:


cibernox:
This improves animation a bit, although not yet fully happy with the solution

Issue #54, opened 2016-03-07
Uncaught TypeError: Cannot read property 'contains' of nullhandleRootMouseDown @ basic-dropdown.js:170


ktrhn:
Power select is wrapped in another component. Everything works fine, but the error is bugging me.

cibernox:
Need more info? EPS Version? Ember version? Browser?

akshaisarma:
I have the same issue with EPS: 0.8.6, Ember: 2.3.0, Chrome: Version 47.0.2526.106 (64-bit)

It basically happens when I bind the onopen action to a handler that causes a DOM change (I insert a textbox next to the power select and component gets rerendered). Regardless of whether I use the select.actions.open() in my onopen handler, any further mouse downs anywhere on the page triggers multiple of those error messages. Also, even if I call select.actions.open(), the dropdown actually doesn't show.

ssured:
Same for me. Got a report in an error logger (Rollbar)

Drills down to this line https://github.com/cibernox/ember-basic-dropdown/blob/master/addon/components/basic-dropdown.js#L193

So either `this.element === null` or `self.document.getElementById(this.get('dropdownId')) === null`. Both situations are easy to guard

I cannot reproduce. It hit me on IE11 on windows, Ember 2.4.1

```
TypeError: Unable to get property 'contains' of undefined or null reference
1
File https://?????/assets/vendor-4e5ea57a00552f990f5f6e1bf478fa46.js line 115234 col 7 in handleRootMouseDown
2
File https://d37gvrvc0wt4s1.cloudfront.net/js/v1.8/rollbar.min.js line 1 col 12510 in e._wrapped
```

cibernox:
It's true that both situations are easy to ward, but they should also be impossible.

I can't tell from memory if something changed lately in this part of the code, but an update might have fixed this. Sounds like the manifestation of a different bug.

Can you first try split this check in two different lines to know which sentence causes the failure? Knowing that might lead to discover the real bug.

ssured:
I'll have an eye on this and see if the bug happens more often; I have seen it once in one week.

Not knowing much of the inner workings of this lib, but `document.getElementById(...)` does rely on the fact that the dom is rendered. Guarding for external influences seems right to me. You are right too, first knowing which of this lines contains the error is a good next step.

cibernox:
Spring cleanup. I'll reopen if this happens again.

PR #63, opened 2016-03-21
Split logic for remote debugging of #54

ssured:


cibernox:
Thanks. That will tell us what went wrong if you get another report.

PR #66, opened 2016-04-04
Bug: jumps when have scrolling left area

Kinjeiro:
fix bug: when activate dropdown in scrolled to left area, window jumps to the left border ($window.scrollLeft() doens't use for left variable)


cibernox:
Can you rebase this?

I understand that this can be hard to automate in tests (like all positioning features in this component), can you make some example in the dummy app (make the body it scrollable maybe?) so in the future is easy to manually test this?

ToMoCoop:
I have a feeling that #79 that I just raised will resolve this issue, so you may not need to re-base...
Worth a check.

Kinjeiro:
@ToMoCoop Thanks for fix, yes, this is that bug.

Issue #67, opened 2016-04-06
Add aria-label to trigger


jmacqueen:
Needed for general WAI-ARIA usage and `ember-power-select`.

PR #68, opened 2016-04-06
67 add aria label to trigger

jmacqueen:
cibernox/ember-basic-dropdown#67

### DIFF
--- a/addon/components/basic-dropdown.js
+++ b/addon/components/basic-dropdown.js
@@ -1,3 +1,4 @@
+/*jshint unused:false*/
 import Ember from 'ember';
 import layout from '../templates/components/basic-dropdown';
 import getOwner from 'ember-getowner-polyfill';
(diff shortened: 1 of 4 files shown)


addon/components/basic-dropdown.js
```diff
@@ -1,3 +1,4 @@
+/*jshint unused:false*/
```
cibernox: Why this? jshint was failing?

cibernox:
LGTM, just one question about a jshint comment.

addon/components/basic-dropdown.js
jmacqueen: 'fraid so. On line [54](https://github.com/jmacqueen/ember-basic-dropdown/blob/67_add_aria-label_to_trigger/addon/components/basic-dropdown.js#L54-L56) `e` is a defined parameter, but not used in the function body.

It's not in any way a critical failure. In fact, it is often a deliberate choice to 'document' that the event is getting passed in for anyone who comes along in the future. So, I opted to tell jsHint that it's okay this time, rather than remove the parameter.

The other alternatives are to remove it entirely and replace `e =>` with `() =>` or, to maintain the self-documenting nature, the rather extreme `(/*e*/) =>` should also work.

Personally, I think it's cleaner and most useful to keep the unused parameter and just let jsHint know it's a deliberate choice this time.

cibernox:
Thanks. I'll release a new version in a moment

PR #70, opened 2016-04-09
Simplified animation logic.

cibernox:
Less run-loops, less requestAnimationFrame, more Ember.

### DIFF
--- a/addon/components/basic-dropdown/content.js
+++ b/addon/components/basic-dropdown/content.js
@@ -30,6 +30,7 @@ function waitForAnimations(element, callback) {
 
 export default WormholeComponent.extend({
   layout,
+  transitionClass: 'ember-basic-dropdown--transitioning-in',
 
   // Lifecycle hooks
   didInsertElement() {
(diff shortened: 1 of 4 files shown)


miguelcobain:
What was the thinking behind these removals?

Seems like magic that the animations can work. :P

cibernox:
@miguelcobain Actually I changed it again recently, but the idea is the same.

Open:

1. When the component opens, the dropdown has a `*--transitioning-in` class.
2. In the `afterRender` I inspect using `window.getComputedStyle` if there is any animation or transition going on.
3.1. If no animation/transition is going on, I swap the `*--transitioning-in` by a `*--transitioned-in`
3.2. If there is, I add an event listener for the `animationend/transitionend` DOM events. That event listener swaps the `*--transitioning-in` by a `*--transitioned-in` class.

Close:

1. I clone the dropdown to create a "ghost" version that I can animate out.
2. Add the `*--transitioning-out` class, check if there is animations:
2.1 - No animation: I remove the clone
2.2 -  Add an event listener for the `animationend/transitionend` DOM events, and once it fires, the "ghost" is destroyed.

I did this because I wanted animations to be as simple to add the CSS, and the JS checks for CSS animations for doing the right thing.

PR #76, opened 2016-04-18
Move initialization logic from didInitAttrs to init hook

davidlormor:
Update to fix deprecation warning: http://emberjs.com/deprecations/v2.x/#toc_ember-component-didinitattrs


cibernox:
Thanks! One less deprecation to go.

PR #79, opened 2016-04-27
Fix horizontal position when using horizontal scroll

ToMoCoop:
When using the dropdown on a page that is wider than 100% of the browser width, scrolling the page horizontal causes the position of the options to be incorrect.

The issue was caused by the left position of the box using just the output of getBoundingClientRect alone, which only accounts for the items position within the browser window and now within the document.
This had been resolved in the vertical scrolling by the addition of triggerTopWithScroll variable, so I have mimiced this with a triggerLeftWithScroll variable.

I think i've kept the logic that detect whether it should draw right or left the same, as I think this should work from the position in the browser and not the document, which is currently what it does.

Tested locally in my project and appears to work as expected.
Hope this helps this odd use case.

### DIFF
--- a/addon/components/basic-dropdown.js
+++ b/addon/components/basic-dropdown.js
@@ -208,7 +208,8 @@ export default Component.extend({
       dropdownHeight, dropdownWidth,                        // dropdown dimensions
       scrollTop, scrollLeft                                 // scroll
     } = this._getPositionInfo(dropdown);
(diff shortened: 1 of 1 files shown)


cibernox:
@ToMoCoop I'll try to review on a flight I have this weekend. If not, I'm back from holidays around 4th.

PR #81, opened 2016-04-29
Tapping should close the dropdown on touch-devices. closes #80

### LINKED ISSUE #80: Touch Device "tapping" to close the dropdown doesn't work
If a dropdown is open, and you tab on any other element, the dropdown won't close as it does on a `click` based event.

This seems to happen because there is no global listener for `touchstart` as there is for `mousedown`.
See [addon/components/basic-dropdown/content.js](https://github.com/cibernox/ember-basic-dropdown/blob/master/addon/components/basic-dropdown/content.js#L35).

Simply adding `this.get('appRoot').addEventListener('touchstart', this.handleRootMouseDown, true);` seems to fix it. If you want I can provide a PR.

Haidy777:
See #80

### DIFF
--- a/addon/components/basic-dropdown/content.js
+++ b/addon/components/basic-dropdown/content.js
@@ -27,12 +27,25 @@ export default WormholeComponent.extend({
   layout,
   mutationObserver: null,
 
+  isTouchDevice: (!!self.window && 'ontouchstart' in self.window),
+
   // Lifecycle hooks
   didInsertElement() {
     this._super(...arguments);
     let dropdown = self.window.document.getElementById(this.get('dropdownId'));
+    let appRoot = this.get('appRoot');
+
     this.handleRootMouseDown = this.handleRootMouseDown.bind(this, dropdown);
-    this.get('appRoot').addEventListener('mousedown', this.handleRootMouseDown, true);
+    this.touchStartHandler = this.touchStartHandler.bind(this);
+    this.touchMoveHandler = this.touchMoveHandler.bind(this);
+
+    appRoot.addEventListener('mousedown', this.handleRootMouseDown, true);
+
+    if(this.get('isTouchDevice')){
+      appRoot.addEventListener('touchstart', this.touchStartHandler, true);
+      appRoot.addEventListener('touchend', this.handleRootMouseDown, true);
+    }
+
     if (!this.get('renderInPlace')) {
       dropdown.addEventListener('focusin', this.get('onFocusIn'));
       dropdown.addEventListener('focusout', this.get('onFocusOut'));
@@ -47,7 +60,15 @@ export default WormholeComponent.extend({
   willDestroyElement() {
     this._super(...arguments);
     let dropdown = self.window.document.getElementById(this.get('dropdownId'));
-    this.get('appRoot').removeEventListener('mousedown', this.handleRootMouseDown, true);
+    let appRoot = this.get('appRoot');
+
+    appRoot.removeEventListener('mousedown', this.handleRootMouseDown, true);
+
+    if(this.get('isTouchDevice')){
+      appRoot.removeEventListener('touchstart', this.touchStartHandler, true);
+      appRoot.removeEventListener('touchend', this.handleRootMouseDown, true);
+    }
+
     this.removeGlobalEvents(dropdown);
     if (this.get('animationEnabled')) {
       this.animateOut(dropdown);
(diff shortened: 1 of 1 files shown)


Haidy777:
@cibernox Unfortunately the build is failing.
https://travis-ci.org/Haidy777/ember-basic-dropdown/jobs/127170225#L295
Any ideas?

ludalex:
Just tried this on iOS 9.3 w/ Safari and although clicking outside of the element works great, when this version of `ember-basic-dropdown` is used for `emeber-power-select`, clicking _inside_ the dropdown content triggers closing as well, making scrolling lists or selecting the search box impossible.

cibernox:
@ludalex I'd expect [this line](https://github.com/cibernox/ember-basic-dropdown/pull/81/files#diff-557380c41041060ba02fda741c1add2bR106) to take care of that. Can you check why it isn't?

ludalex:
Found another little issue as well, always when used in combination with `ember-power-select`:
Basically clicking on the trigger to close the dropdown content works only twice, then stops functioning.
This happens both on mobile and desktop.

ludalex:
@Haidy777 found out what might be the root cause of these nasty bugs:
bugn x
The `handleRootMouseDown` method is called exponentially on every click, probably due to some event registration bug.

Haidy777:
@ludalex Just checked it with the dev tools (getEventListeners(appRoot)).
Seems like it's not removing the listeners which leads to this exponentially thing. I'm trying to figure it out.

Haidy777:
@ludalex Just updated the commit. The exponentially thing looks like it's fixed. Maybe you could test this again?

ludalex:
I can confirm it's fixed!

addon/components/basic-dropdown/content.js
```diff
@@ -117,5 +142,17 @@ export default WormholeComponent.extend({
       dropdown.removeEventListener('DOMNodeInserted', reposition);
       dropdown.removeEventListener('DOMNodeRemoved', reposition);
     }
+  },
+
+  touchMoveHandler() {
+    const appRoot = this.get('appRoot');
+    if (!this.isDestroying && !this.isDestroyed) {
+      this.hasMoved = true;
+      appRoot.removeEventListener('touchmove', this.touchMoveHandler.bind(this));
```
cibernox: This is not detaching the event properly. Every time you use `Function.prototype.bind` your are generating a new function, so this event handler you are removing is not the same event handler you attached a few lines below.

You need to, on init, bind it and store it in a propertly, like you already did with `touchStartHandler`

addon/components/basic-dropdown/content.js
```diff
@@ -74,7 +94,12 @@ export default WormholeComponent.extend({
   },
 
   handleRootMouseDown(dropdownContent, e) {
-    let comesFromInside = this.element.parentElement.contains(e.target) || dropdownContent.contains(e.target);
+    if(this.get('hasMoved')){
```
cibernox: `this.hasMoved` is not a property where we case about observations, so it is save to get/set as any regular property, without using Ember.get/set.

addon/components/basic-dropdown/content.js
```diff
@@ -117,5 +142,17 @@ export default WormholeComponent.extend({
...
+  touchMoveHandler() {
+    const appRoot = this.get('appRoot');
+    if (!this.isDestroying && !this.isDestroyed) {
+      this.hasMoved = true;
+      appRoot.removeEventListener('touchmove', this.touchMoveHandler.bind(this));
+    }
+  },
+
+  touchStartHandler () {
+    this.get('appRoot').addEventListener('touchmove', this.touchMoveHandler.bind(this));
```
cibernox: I think that this event listener should be attached on the capture phase (pass true as the last argument). Same for removing it.

addon/components/basic-dropdown/content.js
```diff
@@ -47,7 +59,15 @@ export default WormholeComponent.extend({
   willDestroyElement() {
     this._super(...arguments);
     let dropdown = self.window.document.getElementById(this.get('dropdownId'));
-    this.get('appRoot').removeEventListener('mousedown', this.handleRootMouseDown, true);
+    const appRoot = this.get('appRoot');
```
cibernox: Use let. I will add `ember-suave` soon and it will compain about const being used outside the top-level scope.

addon/components/basic-dropdown.js
```diff
@@ -163,8 +163,10 @@ export default Component.extend({
     if (!this.get('publicAPI.isOpen')) { return; }
     let onClose = this.get('onClose');
     if (onClose && onClose(this.get('publicAPI'), e) === false) { return; }
-    this.set('publicAPI.isOpen', false);
-    this.setProperties({ _verticalPositionClass: null, _horizontalPositionClass: null });
+    if (!this.isDestroying && !this.isDestroyed) {
```
cibernox: This is something that intrigues me. I don't see any possible situation where this method can be called once the component is destroyed. Did you find one?

cibernox:
I added a few comments, but at a macro scope it looks good 👍

Haidy777:
Thanks. Just updated the commit. I think I got all your comments covered.

cibernox:
Thanks!

PR #82, opened 2016-04-30
Fix IE9 compatibility

BrightOuterSpace:
Currently the component is not working in IE9 due to two problems:

- IE9 does not implement 'classList', therefore adding and removing css-classes is not working. One solution is to use jQuery, another solution could be using a shim, see e.g. https://developer.mozilla.org/en-US/docs/Web/API/Element/classList
- The 'getComputedStyle'-method does not consider 'transitionDuration' in IE9. Therefore 'transitionDuration' is undefined and a transition is always assumed.

### DIFF
--- a/addon/components/basic-dropdown/content.js
+++ b/addon/components/basic-dropdown/content.js
@@ -64,11 +64,11 @@ export default WormholeComponent.extend({
     let parentElement = this.get('renderInPlace') ? dropdown.parentElement.parentElement : dropdown.parentElement;
     let clone = dropdown.cloneNode(true);
     clone.id = clone.id + '--clone';
-    parentElement.appendChild(clone);
-    let $clone = Ember.$('#' + parentElement.id + ' #' + clone.id);
+    let $clone = Ember.$(clone);
     $clone.removeClass('ember-basic-dropdown--transitioned-in');
     $clone.removeClass('ember-basic-dropdown--transitioning-in');
     $clone.addClass('ember-basic-dropdown--transitioning-out');
+    parentElement.appendChild(clone);
     waitForAnimations(clone, function() {
       parentElement.removeChild(clone);
     });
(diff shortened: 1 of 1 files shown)


addon/components/basic-dropdown/content.js
```diff
@@ -64,10 +64,11 @@ export default WormholeComponent.extend({
     let parentElement = this.get('renderInPlace') ? dropdown.parentElement.parentElement : dropdown.parentElement;
     let clone = dropdown.cloneNode(true);
     clone.id = clone.id + '--clone';
-    clone.classList.remove('ember-basic-dropdown--transitioned-in');
-    clone.classList.remove('ember-basic-dropdown--transitioning-in');
-    clone.classList.add('ember-basic-dropdown--transitioning-out');
     parentElement.appendChild(clone);
+    let $clone = Ember.$('#' + parentElement.id + ' #' + clone.id);
```
cibernox: You can pass the clone to jquery directly

```js
let  $clone = $(clone);
```

aaaand if you do so, you should add/remove the classes before appending it to the DOM.

addon/components/basic-dropdown/content.js
```diff
@@ -6,7 +6,7 @@ const { run } = Ember;
 const MutObserver = self.window.MutationObserver || self.window.WebKitMutationObserver;
 function waitForAnimations(element, callback) {
   let computedStyle = self.window.getComputedStyle(element);
-  if (computedStyle.transitionDuration !== '0s') {
+  if (computedStyle.transitionDuration && computedStyle.transitionDuration !== '0s') {
```
cibernox: Do you need a similar check for `animationName`/`animationPlayState`?

cibernox:
Just a minor optimization and a question. I'll merge this tomorrow. I love to know if there is a free/cheap way of running tests in IE9 so we can prevent regressions.

BrightOuterSpace:
Thanks for the feedback! I followed your suggestion, looks way better :)

`animationName` and `animationPlayState` are also undefined in IE9 - but it works because `animationPlayState` is checked for equality. Maybe some more general kind of feature detection would be senseful.

Unfortunately I don't now about a good way for running tests in IE9 - I noticed the problems during manual tests. Automatic tests would be really great.

cibernox:
Tests are red due to an NPM problem. Restarting the build

cibernox:
Thanks!

PR #94, opened 2016-06-02
added a run loop schedule to ensure style modifications come after do…

SpikedKira:
…m manipulation


cibernox:
Thanks for the idea!

PR #101, opened 2016-06-09
Allow to customize the inner components from the outside

cibernox:
@miguelcobain Ping

### DIFF
--- a/addon/components/basic-dropdown.js
+++ b/addon/components/basic-dropdown.js
@@ -5,16 +5,19 @@ import set from  'ember-metal/set';
 import $ from 'jquery';
 import layout from '../templates/components/basic-dropdown';
 import { schedule } from 'ember-runloop';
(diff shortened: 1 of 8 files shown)


cibernox:
I've found a regression in 2.7 that is making this build fail (it was making the build fail before, but it was canary and canary was an allowed failure).

It's not hard to workaround, but I want to check first if it is considered a regression in Ember itself.

miguelcobain:
This is exactly what I needed.

calling `this.elementId` in a tagless component feels strange, but yes, we need to check if it is intended or not.

Thanks.

cibernox:
I just wanted a unique that allows me to relate each trigger with its content, it could be an autoincremental number, but since the element id was there I used it. 

The fix is trivial, but I want to wait to see if this was a breaking change in Ember.

PR #110, opened 2016-06-27
Fix naive reposition window scrollLeft typo

ilucin:


cibernox:
Oops, good catch.

cibernox:
Thanks! I'll release a new version now

ilucin:
Yep, [this](http://take.ms/t3WMx) looked kinda wrong to me :D 

Tnx for quick reaction!

PR #115, opened 2016-07-07
Add fastboot-dist to npmignore

jakesjews:
Its not alot of space but this reduces the install size by 2.7MB


cibernox:
Every little helps.

Thanks

Issue #105, opened 2016-06-14
 TypeError: Cannot read property 'querySelector' of undefined in FastBoot


josemarluedke:
After fixing cibernox/ember-power-select#543 this bug happens when running on fastboot.

`TypeError: Cannot read property 'querySelector' of undefined`

I suspect that this is happening here: https://github.com/cibernox/ember-basic-dropdown/blob/fa5949ad1940893abbac02fa5f5929cc256d9bab/addon/components/basic-dropdown.js#L55

josemarluedke:
I hope to get #116 merged with the fix, mean while I have published my test app with the reproduction of the error at https://github.com/josemarluedke/test-app-fastboot!

Thanks @Rastopyr for stepping in and fixing the problem.

Rastopyr:
@josemarluedke np! 👍

Rastopyr:
@cibernox we are fixed API for correct FastBoot rendering, so will be any updates for publishing new version of `ember-basic-dropdown` in npm registry?

i think is very matter for correct work of `ember-power-select`.

cibernox:
Published a new version. Given that the dependency of EPS in EBD is loose, deleting EBD in your node_modules folder and install again should be enough

PR #116, opened 2016-07-11
[Fix] Add missed checking for fastboot rendering

Rastopyr:


package.json
```diff
@@ -1,6 +1,6 @@
 {
   "name": "ember-basic-dropdown",
-  "version": "0.12.0-beta.22",
```
cibernox: can you revert this line? It is incremented automatically what I run `ember release`

Rastopyr: Done :)

Rastopyr:
Revert full commit, because have lint-error by jscs.
And, i think, we should link this PR with #105 issue.

cibernox:
This week I won't be able to do do any OS, but I'll review all pending PRs over the weekend.

cibernox:
Thanks

Rastopyr:
np 👍

PR #120, opened 2016-07-25
[WIP] Immutable api

cibernox:
Treat the publicAPI as an immutable object. Every change replaces the API with a new object. When changes happen, once per runloop in the actions queue, the `registerAPI` action is called so changes are propagated.


cibernox:
Ember Power Select works with this approach, so I'm going to merge 🎉

PR #135, opened 2016-08-17
Close on click not on dropdown

cibernox:


cibernox:
This ended up being a bad idea.

Issue #130, opened 2016-08-04
removeEventListener exception on IE 11


Kilowhisky:
Get this error when navigating away from a route with a dropdown in IE 11 when i have the developer tools open and i run a production build. Can't be sure it happens when the tools are closed and it doesn't happen when using a development build.  

Dropdown declaration. 

```
{{#power-select-multiple renderInPlace=true options=assets selected=filteredAssets searchField="name" placeholder=(t 'track.trackEventsChooseAssetsToShow') onchange=(action "assetFilterChanged") as |asset|}}
              {{icon-asset asset=asset class="text-bottom" width="20px" height="20px"}}
              {{asset.name}}
              {{#if asset.registration}}<div class="description">({{asset.registration}})</div>{{/if}}
{{/power-select-multiple}}
```
Stack Trace
```
"TypeError: Unable to get property 'removeEventListener' of undefined or null reference
   at removeGlobalEvents (https://myapp/assets/vendor-d42d5df2bd86ce26959f3ab73da5b5bd.js:34:23452)
   at _teardown (https://myapp/assets/vendor-d42d5df2bd86ce26959f3ab73da5b5bd.js:34:24415)
   at willDestroyElement (https://myapp/assets/vendor-d42d5df2bd86ce26959f3ab73da5b5bd.js:34:20486)
   at r (https://myapp/assets/vendor-d42d5df2bd86ce26959f3ab73da5b5bd.js:11:30486)
   at trigger (https://myapp/assets/vendor-d42d5df2bd86ce26959f3ab73da5b5bd.js:16:814)
   at r (https://myapp/assets/vendor-d42d5df2bd86ce26959f3ab73da5b5bd.js:11:30486)
   at d.prototype.willDestroyElement (https://myapp/assets/vendor-d42d5df2bd86ce26959f3ab73da5b5bd.js:9:10395)
   at d.prototype.remove (https://myapp/assets/vendor-d42d5df2bd86ce26959f3ab73da5b5bd.js:"
```

cibernox:
That is weird, because looking at the stack trace seems that this error comes from `removeGlobalEvents`, and in that method the events are being removed from `window`. There is no way that window can be undefined 😲

cibernox:
Oh, I see that this method does remove events from `dropdownElement`. However that's only the case if the browser doesn't support mutation observer. IE11 supports MutationObserver, according to can-i-use.

Kilowhisky:
[...]

PR #145, opened 2016-09-01
Fix issue where this.dropdownElement is undefined on IE 10

jdurand:
fixes https://github.com/cibernox/ember-basic-dropdown/issues/130


cibernox:
Yes this makes sense 👍

cibernox:
Thanks! Just published a new version.

PR #146, opened 2016-09-02
tabIndex computed needs to depend on tabindex attr

xomaczar:


xomaczar:
@cibernox did you have a chance to look at this PR

PR #157, opened 2016-09-22
updates ember-get-config dependency

noslouch:
This all started as in investigation into [ember-power-select](https://github.com/cibernox/ember-power-select) error that was thrown from [ember-get-config](https://github.com/null-null-null/ember-get-config).

The nature of the original problem is fixed by recent updates to the `ember-get-config` lib. It used to use the `window` for lookups, but has made updates recently for `fastboot` and other contexts where available objects were not defined on the `window`.

This PR bumps the version number of `ember-get-config` to its latest and adds a required setup line to the `index.js` file.

### DIFF
--- a/index.js
+++ b/index.js
@@ -8,6 +8,10 @@ module.exports = {
     var app = appOrAddon.app || appOrAddon;
     if (!app.__emberBasicDropdownIncludedInvoked) {
       app.__emberBasicDropdownIncludedInvoked = true;
+      // for ember-get-config compat
+      // https://github.com/null-null-null/ember-get-config#usage
+      this.eachAddonInvoke('included', [app]);
+      
       this._super.included.apply(this, arguments);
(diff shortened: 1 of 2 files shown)


noslouch:
^^^ this failure seems like a cache dependency. `stringify-object` is a dependency of bower and (seemingly) unrelated to `ember-get-config`

cibernox:
@noslouch I don't really like the idea of call the included hook of all the addons of the app like that addon suggests. 

Can you replace the `this.eachAddonInvoke('included', [app])` by this?
```
      var emberGetConfig = this.addons.filter(function(addon) {
        return addon.name === 'ember-get-config';
      })[0];
      emberGetConfig.included.apply(emberGetConfig, [app]);
```

This should only call the included hook of the addon we care about. Maybe I'm being paranoid, but better safe than sorry.

PR #160, opened 2016-09-25
Update ember-get-config to a version that works in fastboot

cibernox:
Closes #157

Since this issue (https://github.com/null-null-null/ember-get-config/issues/10), no calls special magic is required

### DIFF
--- a/package.json
+++ b/package.json
@@ -59,7 +59,7 @@
   "dependencies": {
     "ember-cli-babel": "^5.1.10",
     "ember-cli-htmlbars": "^1.1.0",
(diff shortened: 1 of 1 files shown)


noslouch:
Rad! Thanks!

Could I bother you to update ember-power-select to use an updated version of ember-basic-dropdown?

cibernox:
No need actually. The dependency of EPS in EBD is relaxed. If you wipe your node_modules it will get the new version.

noslouch:
Heads up @wismer. That issue we were seeing is fixed. Thanks @cibernox!

PR #171, opened 2016-11-11
Fixed issues in calculate-position.js (support for horizontal scrolling primarily)

sebastianseilund:
Thanks for ember-basic-dropdown and ember-power-select!

There are some positioning issues, which makes the dropdown unusable, when the viewport is horizontally scrolled. Not scrolled:

screen shot 2016-11-11 at 14 20 16

Scrolled:

screen shot 2016-11-11 at 14 20 24

This PR fixes that and some more positioning issues. This is what I changed:

- Added support for when the page is horizontally scrolled. The function wasn't taking
  `scroll.left` into account, which causes the dropdown to drift to the
left as you scroll right.
- Improved horizontalPosition=auto to work better for dropdowns that are
  smaller than their trigger element. It now considers which position
will show the most horizontal pixels of the dropdown in either left- or
right-aligned mode and updates the position accordingly.
- Made sure that `style.left` is set to `null` when `style.right` gets
  set. The dropdown element should only have a `left` or a `right` css
property at any time.
- Reorganized the algorithm a little. No need to set `dropdownLeft` fx when
  we end up setting `style.right`. Also added some comments about the
parts that are less obvious/hard to grok.

This also fixes the issues described in #79 and #66, but they don't seem to take all the `horizontalPosition` options into account, so `horizontalPosition=right` still doesn't work with horizontal scrolling.

Let me know if you want me to change anything.

### DIFF
--- a/addon/utils/calculate-position.js
+++ b/addon/utils/calculate-position.js
@@ -17,38 +17,56 @@ import $ from 'jquery';
     - {Object} CSS properties to be set on the dropdown. It supports `top`, `left`, `right` and `width`.
 */
 export function calculatePosition(trigger, dropdown, { previousHorizontalPosition, horizontalPosition, previousVerticalPosition, verticalPosition, matchTriggerWidth }) {
+  // Collect information about all the involved DOM elements
   let $window = $(self.window);
   let scroll = { left: $window.scrollLeft(), top: $window.scrollTop() };
   let { left: triggerLeft, top: triggerTop, width: triggerWidth, height: triggerHeight } = trigger.getBoundingClientRect();
   let { height: dropdownHeight, width: dropdownWidth } = dropdown.getBoundingClientRect();
-  let dropdownLeft = triggerLeft;
-  let dropdownTop;
+  let viewportWidth = self.window.innerWidth;
+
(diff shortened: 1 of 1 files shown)


cibernox:
Hey, this is fantastic. Could you just fix the ESLint warnings? Tomorrow I'm going  to test this logic in a variety of situations. I haven't found a good way of testing this automatically, so I'll be the human QA on this, but changes I've seen make total sense.

sebastianseilund:
Thanks. I fixed the ESLint warnings.

Sounds good!

I already tested it manually with all all combinations of `horizontalPosition=auto|left|center|right` and the dropdown being narrower and wider than the trigger. But more testing wouldn't hurt :)

cibernox:
I've tested it and it looks good. 

I've decided that there is no need in setting left/right to null. That logic should be in the function that applies the styles. Left and right are mutually exclusive. I fixed it on master, but there is no need to change anything here.

cibernox:
Thanks! Positioning is much better and clearer now. I'll  release a new version soon. EPS has a relaxed dependency in EBD, so just reinstalling your packages should get you the fix.

PR #187, opened 2016-12-22
call calculatePosition with context set to the component

miguelcobain:
We'd like to use eps `calculatePosition` hook in ember-paper. What is stopping us from that is that we need access to the component to get some options the user can provide.
We've done that through the function's context.

You probably don't want to leak internal stuff into public APIs, but this could be a solution for our problem.

### DIFF
--- a/addon/components/basic-dropdown.js
+++ b/addon/components/basic-dropdown.js
@@ -148,7 +148,7 @@ export default Component.extend({
 
     let calculatePosition = this.get(this.get('renderInPlace') ? 'calculateInPlacePosition' : 'calculatePosition');
     let options = this.getProperties('horizontalPosition', 'verticalPosition', 'matchTriggerWidth', 'previousHorizontalPosition', 'previousVerticalPosition');
-    let positionData = calculatePosition(triggerElement, dropdownElement, options);
+    let positionData = calculatePosition.call(this, triggerElement, dropdownElement, options);
     return this.applyReposition(triggerElement, dropdownElement, positionData);
   },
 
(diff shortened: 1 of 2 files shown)


miguelcobain:
As a workaround I'm using a computed property like:

```js
calculatePosition: computed(function() {
  return this.calculatePositionAux.bind(this);
}),
```

cibernox:
The problem with setting the context of the function is that then you cannot use **your** context like this:

```
{{basic-dropdown calculatePosition=(action 'my-position-strategy')}}
```

where `my-position-strategy` expects `this` to be the outer controller/component. I'd rather pass the component itself as one of the options in the object (https://github.com/cibernox/ember-basic-dropdown/blob/master/addon/utils/calculate-position.js#L19) and allow the context to be the usual one expected by the user.

Besides, this approach would be less breaking.

Other than that this seems a reasonable idea.

miguelcobain:
@cibernox what I don't understand is why does the context gets erased if I define a `calculatePosition` in a subclass.

cibernox:
I think that it's just how javascript works. If a function is extracted of it's context, it doesn't retain it, unless it is bound. By example: 
screen shot 2016-12-22 at 14 54 24

miguelcobain:
The thing is that I'm not taking it out of the context myself. I'm basically doing:

```js
export default BasicDropdownComponent.extend({
  calculatePosition() {
    // this doesn't point to the current component
    // which feels weird
  }
})
```

cibernox:
It is the addon itself the one that extracts it: https://github.com/cibernox/ember-basic-dropdown/blob/master/addon/components/basic-dropdown.js#L149-L151

That is why I was suggesting to pass the component itself in the arguments, so people don't rely on `this`. If we don't bind the context of the function to anything, users can still bind it to, by example, their controller.

miguelcobain:
I don't see how

```js
let func = this.get('func');
func();
```

is taking it out of context.

Is it that different from
```
this.func();
```
?

cibernox:
@miguelcobain Yes, entirely different. Check the snapshot from the jsbin above.

This is one of those "JS is 4W3S0M3!!!1!" moments 🤦‍♂️

PR #189, opened 2016-12-23
Update ember-cli-sass

xomaczar:


xomaczar:
Yeah I know, he did not check the latest code changes - I've raised that issue with the repo owner. What changed: support for node-sass 4.1.0 which now compiles on alphine linux - which ember-twiddle is running. So hopefully, we can use ember-basic-dropdown and ember-power-select with ember-twiddle.

cibernox:
if you rebase this should be green too

xomaczar:
Will do. Does ember-power-select suffers from the same issue?

cibernox:
i fixed both

xomaczar:
All green.

PR #191, opened 2016-12-25
Remove node-sass dev dependency

xomaczar:


cibernox:
Published new versions of everything.

xomaczar:
Thx. Merry Xmas to u too

PR #192, opened 2017-01-04
Permit overrides of default handleMousedown, handleKeydown, handleTouchend actions

jerwilkins:
When using the Ember Basic Dropdown with custom events, it is sometimes desirable (in my case necessary) to override the default event handlers, particularly when working with the `mouseenter`/`mouseleave` optional handlers. 

Also added `dropdown` as a parameter for the default handlers (and their bound overrides, if applicable) for consistency with the optional handlers. 

Added tests for the new functionality, and aimed for consistency with existing test suite and application code. 

Thanks for the great project!

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -18,3 +18,4 @@ npm-debug.log*
 testem.log
 .vscode
 .env.deploy.*
+/.idea
--- a/addon/components/basic-dropdown/trigger.js
+++ b/addon/components/basic-dropdown/trigger.js
@@ -115,6 +115,12 @@ export default Component.extend({
         return;
       }
       this.stopTextSelectionUntilMouseup();
+      // execute user-supplied onMousedown function before default toggle action;
+      // short-circuit default behavior if user-supplied function returns `false`
+      let onMousedown = this.get('onMousedown');
+      if (onMousedown && onMousedown(dropdown, e) === false) {
+        return;
+      }
       dropdown.actions.toggle(e);
     },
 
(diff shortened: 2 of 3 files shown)


cibernox:
I'm curious to know what kind of behaviour you want to achieve with those actions that cannot be done today.

If this is merged, I'd rather take a *slightly* different approach. Almost a naming change. I'd prefer the user to provide `onMouseDown`, `onKeyDown` and `onTouchEnd` actions. Thos actions, if provided, are called before the default behaviour. They don't override the default behaviour unless the user return `false` from them. This makes easier for user to "suscribe" to those events but not have to replicate the default behaviour.

jerwilkins:
The desired behavior (from both this PR and the successive one) is to facilitate a hover-style dropdown menu, where the default events for toggling the dropdown are no-ops, and where mouseenter/leave and focusin/out events are captured at the dropdown level (rather than at the dropdown.trigger level).

I think what you've proposed would be a great approach! Mind if I give it a shot?

cibernox:
Thanks! I'll release a new version in a minute

cibernox:
@jerwilkins I've just realised that you named the action `onMousedown` and `onTouchend`, but the other actions received in the component are more camelcased (`onMouseEnter`/`onMouseLeave`/`onFocusOut`...).

Can you create another PR renaming the new actions to be consistent?

jerwilkins:
Beat me to it, looks like ;) 

One thing I was wondering is if `onKeydown` should become `onKeyDown` as well, but that override option has been there for a long time, so it could be a breaking change.

cibernox:
I think it should. I can still afford breaking changed as long as there is a bump in the middle digit.

PR #195, opened 2017-01-09
[FEATURE] Allow to add mouseenter/leave events to the dropdown content

cibernox:


cibernox:
@jerwilkins Check this out. I've added the primitives to add mouseenter/leave events to the content, and used that to hack an example in the dummy app: https://ember-basic-dropdown.pagefrontapp.com/

(last example. The trigger can still be clicked but you'll fix that on your pr)

What do you think. A bit low level, but it works even when the content is rendered in the root of the body, while your approach requires `renderInPlace=true` to work.

tests/dummy/app/controllers/index.js
```diff
@@ -33,6 +35,38 @@ export default Ember.Controller.extend({
...
+      if (this.closeTimer) {
+        run.cancel(this.closeTimer);
+        this.closeTimer = null;
+      } else {
+        dropdown.actions.open();
+      }
+    },
+
+    triggerMouseLeave(dropdown /*, e */) {
+      this.closeTimer = run.next(this, function() {
```
stefanpenner: you should likely also cancel any current `closeTimer` before setting a new one, that way the system cannot accidentally orphan a timer.

cibernox: @stefanpenner How did you come across this code that doesn't even exist today and was on the dummy app?

jerwilkins: I was showing @stefanpenner some old PRs I did on ember-basic-dropdown and the eventual solution you landed on for hover support.

`3. Miguel’s eventual approach: https://github.com/cibernox/ember-basic-dropdown/pull/195/files`

Nothing actionable here 😅

cibernox: All right Sir. Carry on! 😄

PR #197, opened 2017-01-10
Update README with body height info (closes #196)

### LINKED ISSUE #196: Body height is required to trigger close on click
I'm working on a new app with just a header, so the body has a height of just 40px. Clicking anywhere but on the header (aka `body`) would fail to close the modal.

A simple css line fixed that: `body {min-height: 100vh;}`. 

It is unlikely to happen in an actual app and not a bug per say but I thought I would share anyway. Maybe it could be added to the docs.

karellm:
Details at #196

### DIFF
--- a/README.md
+++ b/README.md
@@ -76,6 +76,8 @@ You can make the dropdown content standout a little more by adding `overlay=true
     {{/basic-dropdown}}
 ```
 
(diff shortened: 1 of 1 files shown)


cibernox:
Thanks!

PR #200, opened 2017-01-11
Bugfix/consistent event handler case

jerwilkins:
Remaining event handler overrides, as well as the trigger actions, moved to camel case


cibernox:
Thanks!

PR #210, opened 2017-01-16
Fix typo in 'public-pages/cookbook/index.hbs'

loboulet:
`Comming` → `Coming`


loboulet:
Branch name is not accurate. Will reopen with a better one 👯‍♂️

Issue #218, opened 2017-02-02
[a11y] Refine aria implementation


steveszc:
These items will address some a11y issues outlined in [ember-power-select issue #806](https://github.com/cibernox/ember-power-select/issues/806) that are specifically related to the basic-dropdown component. 

- [x] Use aria-owns instead of aria-controls to associate the combobox with the listbox<details>
 https://www.w3.org/TR/wai-aria/roles#combobox
</details>

- [x]  aria-owns/aria-controls must reference an element that exists in the dom.<details>
Currently the trigger's `aria-controls` attribute references an ID of a dom node that does not exist when the power-select is in the collapsed state. This is considered invalid markup and can break screen readers. The controlled element (the listbox) should be maintained in the dom even when collapsed, it can be an empty `div`/`ul`. - https://www.w3.org/TR/wai-aria/states_and_properties#aria-owns
</details>

- [x]  aria-haspopup should not be used on the trigger<details>
`aria-haspopup` is an implicit property of a combobox. Explicitly adding aria-haspopup implicitly and incorrectly defines the element as an aria menu (similar in semantics to a right-click contextual menu or the menu ribbon in MS Word or an OS), which not the correct role for the trigger. This attribute is not necessary on a combobox. 
https://www.w3.org/TR/wai-aria/states_and_properties#aria-haspopup 
https://www.w3.org/TR/wai-aria/roles#combobox
</details>

- [x]  aria-pressed should not be used on the trigger.<details>
`aria-expanded` is sufficient to imply behavior to screen readers. `aria-pressed` is intended for toggle buttons such as checkboxes or tri-state buttons.
https://www.w3.org/TR/wai-aria/states_and_properties#aria-pressed
</details>

cibernox:
> Use aria-owns instead of aria-controls

steveszc:
Per the w3c wai-aria documentation included above, `aria-haspopup` and `aria-pressed` mean something different to a screen reader than a button that expands/collapses some content. For the purposes of basic-dropdown it is sufficient to use `aria-expanded` to indicate the expanded state of the component, and use `aria-owns` to create a parent/child relationship between the two dom nodes (trigger and content). `aria-owns` carries less implicit meaning than `aria-controls` so it's probably a better default for a generic expand/collapse component like basic-dropdown.

cibernox:
Perfect, I'll apply at least some of this feedback this weekend.

cibernox:
[...]

PR #224, opened 2017-02-09
Have a constant element in DOM

cibernox:
Part of #218 

/cc @steveszc can you verify that this makes a difference and the fact the the div moves from beneath the input to the root of the body is not a problem? Render the div in the wormhole has a couple issues. Because it some glimmer2 limitations having an IF right inside an ember-wormhole, without a stable parent, causes a render problem (well know issue).

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,3 +1,6 @@
+- [BREAKING] It is a problem for a11y to have `aria-owns/controls` to an element that it's not
+  in the DOM, so now there is a stable div with the right ID that gets moved to the root 
+  of the body when the component opens.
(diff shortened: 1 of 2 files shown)


cibernox:
You might want to use https://github.com/cibernox/ember-power-select/pull/811 and `npm link` this branch of ember-basic-dropdown if you want to test it along with EPS. There is still work to do in EPS tho.

steveszc:
This addresses the issue of aria-owns referencing a dom node that doesn't exist, the node's location in the dom shouldn't matter at all. I'd say this fixes the issue.

cibernox:
Great, I'll merge this as soon as I'm sure it doesn't cause any problem in EPS and proceed to with the a11y feedback in it.

PR #228, opened 2017-02-13
Fix touch handler synthetic click

ilucin:
If `e.target` is an SVG element `click()` method doesn't exist. jQuery takes care of that.

### DIFF
--- a/test-support/helpers/ember-basic-dropdown.js
+++ b/test-support/helpers/ember-basic-dropdown.js
@@ -63,7 +63,7 @@ export function clickTrigger(scope, options = {}) {
 }
 
 export function tapTrigger(scope, options = {}) {
-  let selector = `.ember-basic-dropdown-trigger ${options.triggerChildSelector || ''}`.trim();
+  let selector = '.ember-basic-dropdown-trigger';
   if (scope) {
     selector = scope + ' ' + selector;
   }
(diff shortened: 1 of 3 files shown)


cibernox:
Nice edge case!
Can you add a regression test that covers this?

PS: I'm curious about what are you building with SVG

ilucin:
> Nice edge case!

Yep. Good thing is the `click` method is invoked asynchronously so the error doesn't stop current event loop stack and everything still mostly works.

> Can you add a regression test that covers this?

Here it is. The test will fail if the fix is reverted.

> PS: I'm curious about what are you building with SVG

Nothing special, it's just an SVG icon inside of the trigger. This one in particular: http://take.ms/TeWjm :)

test-support/helpers/ember-basic-dropdown.js
```diff
@@ -54,7 +54,7 @@ export function clickTrigger(scope, options = {}) {
 }
 
 export function tapTrigger(scope, options = {}) {
-  let selector = '.ember-basic-dropdown-trigger';
+  let selector = `.ember-basic-dropdown-trigger ${options.triggerChildSelector || ''}`.trim();
```
cibernox: Instead of defining this ad-hoc option, you can use the `nativeTap` function I've just added to trigger a tap in the element inside.

cibernox: Rebase over master to get that helper.

cibernox:
Approved. Let's wait for travis.

cibernox:
Thanks!

PR #232, opened 2017-02-18
[Fixes #231] Move ember-native-dom-helpers to dependencies

### LINKED ISSUE #231: Should ember-native-dom-helpers be in dependencies, rather than devDependencies?
I was including `clickTrigger` in a component integration test like this:

```javascript
import { clickTrigger } from '../../helpers/ember-basic-dropdown';
```

When I upgraded from v0.24.2 to v0.24.4, I got the error:

```
Error: Could not find module `ember-native-dom-helpers/test-support/helpers` imported from `tycho/tests/helpers/ember-basic-dropdown`
TypeError: Cannot read property 'exports' of undefined at http://localhost:7357/assets/test-support.js, line 5774
```

ember-native-dom-helpers isn't getting installed via yarn because it's in devDependencies. 

Should this library be included in dependencies so that users of the test helpers don't have to install the library manually? Or am I including `clickTrigger` incorrectly?

paulcwatts:
I verified this causes ember-native-dom-helpers to be added to node_modules in my app.

### DIFF
--- a/package.json
+++ b/package.json
@@ -49,7 +49,6 @@
     "ember-font-awesome": "^3.0.2",
     "ember-href-to": "1.11.0",
     "ember-load-initializers": "^0.6.0",
-    "ember-native-dom-helpers": "^0.1.3",
     "ember-pagefront": "0.11.2",
     "ember-resolver": "^2.0.3",
     "ember-source": "^2.12.0-beta.1",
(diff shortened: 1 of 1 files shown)


cibernox:
Thanks!

cibernox:
I'll release a new version on Monday or so.

PR #233, opened 2017-02-21
Positioning is off when offset parent is positioned

ef4:
This PR demonstrates how to break positioning of the basic dropdown.

screen shot 2017-02-21 at 10 51 58 am

In short, we can't always assume the wormhole target will be at the viewport's `{ top: 0, left: 0 }`.

I think the most bulletproof way to address this class of bug is to use `getBoundingClientRect` on the wormhole target, and subtract those dimensions from the dropdown content's position. That rules out a whole bunch of otherwise annoying possibilities, like margin collapse percolating up from deeply nested children, etc.

I am unsure of how you'd prefer to do this without impacting public API. Right now the position calculations happen in `basic-dropdown`, but only `basic-dropdown/content` knows the destination ID.

### DIFF
--- a/tests/dummy/app/styles/_base.scss
+++ b/tests/dummy/app/styles/_base.scss
@@ -16,6 +16,8 @@ html {
 
 body {
   color: $text-color;
+  position: relative;
+  margin-top: 60px;
 }
 
 a {


cibernox:
I think that what you suggest is right. The default algorithm should account for any possible displacement of the destination. Another place where this shows is in tests, when the dropdown is inside the small qunit window.

I don't think implementing this will break the API, but I'll check it.

PR #238, opened 2017-02-27
Unify `calculatePosition` and `calculateInPlacePosition`

cibernox:
/cc @miguelcobain


miguelcobain:
This LGTM. It is what i had in mind. 👍

PR #239, opened 2017-02-27
allow calculatePosition to return custom height

miguelcobain:
Followed the same path as `width`.


cibernox:
Merged and released on 0.30.1

Issue #262, opened 2017-04-11
horizontalPosition='right' leaves an offset


mydea:
When using `horizontalPosition='right'`, or when it auto-positions accordingly, it leaves some space for me on the right. You can see this on the demo, when choosing 'Horizontal: right': https://www.ember-basic-dropdown.com/docs/position:

![image](https://cloud.githubusercontent.com/assets/2411343/24915867/a7b22d8e-1ed8-11e7-95b7-11d4914ef736.png)

cibernox:
<img width="424" alt="screen shot 2017-04-11 at 16 08 19" src="https://cloud.githubusercontent.com/assets/265339/24916111/23cfdaf4-1ed1-11e7-9751-527c2bae3ad7.png">


That's weird, I just tried and looked perfect for me. More details? Browser? Screen size?

mydea:
I'm on Windows 10, tried in both Chrome as well as Firefox. 

Now, I think I know what the problem is: It struggles with the scroll bar! If the scroll bar is not visible (= the site content is not too long), the positioning works. But if the scroll bar is there, it is off by the width of the scroll bar. On Mac, the scroll bar doesn't take any width of the screen, so that's the difference there.

mydea:
After poking around a bit more (https://developer.mozilla.org/en-US/docs/Web/API/Element/clientWidth), I think that might fix it, in https://github.com/cibernox/ember-basic-dropdown/blob/master/addon/utils/calculate-position.js#L32:

```js
 let viewportWidth = document.body.clientWidth || self.window.innerWidth;
```

cibernox:
Good catch really. I almost forgot that scrollbars are a thing.

Could you test that make a PR with that? I'm busy today preparing a talk for this thursday meetup.

PR #263, opened 2017-04-12
Fix calculation of position to take scroll bars into account

mydea:
This fixes https://github.com/cibernox/ember-basic-dropdown/issues/262, by using `document.body.clientWidth` instead of `window.innerWidth` to calculate the screen width.


cibernox:
Good catch. Thanks! 🙇

PR #268, opened 2017-04-18
Listen for scrolling on user defined elements

### LINKED ISSUE #267: Cannot handle scrolling that is not on the window
If the dropdown is inside an element that is scrollable the event listeners will not pick up the scroll event. It would be great if the user could pass [a string containing one or more CSS selectors separated by commas](https://developer.mozilla.org/en-US/docs/Web/API/Document/querySelectorAll) which would then be added to the global listeners for the `scroll` event.

non-scrolling_dropdown

RustyToms:
Adds the `scrollableElements` property which is  an optional [string containing one or more CSS selectors separated by commas](https://developer.mozilla.org/en-US/docs/Web/API/Document/querySelectorAll) which are added to the global listeners for the `scroll` event.

closes #267

Before:
```hbs
{{#basic-dropdown horizontalPosition="right" as |dropdown|}}
  {{#dropdown.trigger classNameBindings=":helix-dropdown dropdown.isOpen:open"}}
  <label class="helix-dropdown__options-label ttable__options-label {{if dropdown.isOpen "helix-dropdown__toggler--open"}}">
    <i class="fa fa-cog helix-dropdown__options-cog"></i>
  </label>
  {{/dropdown.trigger}}
  {{#dropdown.content}}
    ...
  {{/dropdown.content}}
{{/basic-dropdown}}
```
non-scrolling_dropdown

After:
```hbs
{{#basic-dropdown horizontalPosition="right" scrollableElements=".content-window" as |dropdown|}}
  {{#dropdown.trigger classNameBindings=":helix-dropdown dropdown.isOpen:open"}}
  <label class="helix-dropdown__options-label ttable__options-label {{if dropdown.isOpen "helix-dropdown__toggler--open"}}">
    <i class="fa fa-cog helix-dropdown__options-cog"></i>
  </label>
  {{/dropdown.trigger}}
  {{#dropdown.content}}
    ...
  {{/dropdown.content}}
{{/basic-dropdown}}
```
scrolling_dropdown

### DIFF
--- a/addon/components/basic-dropdown.js
+++ b/addon/components/basic-dropdown.js
@@ -40,6 +40,7 @@ export default Component.extend({
   right: null,
   width: null,
   height: null,
+  scrollableElements: '',
 
   // Lifecycle hooks
   init() {
@@ -53,6 +54,7 @@ export default Component.extend({
       uniqueId: guidFor(this),
       isOpen: this.get('initiallyOpened') || false,
       disabled: this.get('disabled') || false,
+      scrollables: this.get('scrollableElements'),
       actions: {
         open: this.open.bind(this),
         close: this.close.bind(this),
--- a/addon/components/basic-dropdown/content.js
+++ b/addon/components/basic-dropdown/content.js
@@ -210,7 +210,7 @@ export default Component.extend({
 
   removeGlobalEvents() {
     let scrollables = this.get('scrollingGlobals');
-    if (scrollables.length) {
+    if (scrollables && scrollables.length) {
       scrollables.forEach(element => {
         element.removeEventListener('scroll', this.runloopAwareReposition);
       });
--- a/addon/components/basic-dropdown/content.js
+++ b/addon/components/basic-dropdown/content.js
@@ -40,6 +40,7 @@ export default Component.extend({
   isTouchDevice: (!!self.window && 'ontouchstart' in self.window),
   hasMoved: false,
   animationClass: '',
+  scrollingGlobals: null,
   transitioningInClass: 'ember-basic-dropdown--transitioning-in',
   transitionedInClass: 'ember-basic-dropdown--transitioned-in',
   transitioningOutClass: 'ember-basic-dropdown--transitioning-out',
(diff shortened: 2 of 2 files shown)


RustyToms:
I didn't update the docs or anything because I wasn't sure that this is the approach you would like to take (assuming you would like to accept this). Let me know if you'd like me to refactor this, or update anything else.

cibernox:
I think I have a different approach for situations like this that might work for you. 
If you put the placeholder for the dropdown *inside* the scrollable element, you can tell the dropdown to render in that placeholder. 

E.g:

```hbs
<div class="scrollable-container">
  <div id="basic-dropdown-target-inside-scrollable-container"></div>
  <!-- Blah blah blah content -->
  {{#basic-dropdown horizontalPosition="right" as |dropdown|}}
  {{#dropdown.trigger}}Click me{{/dropdown.trigger}}
  {{#dropdown.content to="basic-dropdown-target-inside-scrollable-container"}}
    ...
  {{/dropdown.content}}
{{/basic-dropdown}}
</div>
```

Can you confirm if this works? Your solution is ingenious tho. If this idea doesn't work something like your idea could be good, although I'd like the component to be smart enough to automatically find the closest scrollable parent without you telling it which one.

RustyToms:
Sure @cibernox thanks, I am working on that now. I also just pushed a small commit that fixes a problem with removing the listeners if there are no scrollableElements defined.

RustyToms:
I set the dropdown content to go inside the scroll parent, and I set the target container to be position: relative, but the positioning still doesn't seem to work correctly. The scrolling works fine.
position dropdown inside scroll parent
Perhaps there is a simple fix for that, I'm not sure. We have run into this problem before, and used a similar solution.

The real problem is when you have a scrolling container inside another scrolling container. The contradicting requirements that the content of the container not overflow (but scroll instead) while the tooltip itself is supposed to overflow has been giving me headaches for a long time. I'll make another gif to show what I mean. 
scrolling_container
This is actually the main reason I was excited to use your dropdown, because we need to be able to position a dropdown outside of its container, but the container itself might be scrollable, and I didn't want to write all of our own code for this. Your repo is really really close (really nice code by the way, I found this because of your entertaining talk at Ember Conf).

The solution in this PR, or perhaps something like it, allows the dev to identify multiple scrolling elements and watch them all. What do you think? Anything else I should try?

cibernox:
The main issue with your approach, albeit small, is that every scroll event has to reposition the dropdown because the container is the body. You can see that there is like a small delay in the position while you scroll. 
If the element is inserted inside the scrollable element it doesn't need to recalculate the position with every scroll because it origin of coordinates is inside the scrollable area. Less work and no delay.

Could you add a reproduction example in the dummy app inside this very repo? Create a route if you have to. It's an interesting problem to solve actually.

RustyToms:
Sure, it will take me some time to become familiar enough with the repo, I won't be able to do it today. I will let you know when I have done it.

PR #272, opened 2017-04-21
Documentation update

ksin:
Some small documentation updates. 

Included babel polyfill because tests were giving `regeneratorRuntime is not defined ember` errors.

### DIFF
--- a/README.md
+++ b/README.md
@@ -1,5 +1,7 @@
 # Ember-basic-dropdown
 
+[![Build Status](https://travis-ci.org/cibernox/ember-basic-dropdown.svg?branch=master)](https://travis-ci.org/cibernox/ember-basic-dropdown)
+
 This is a very minimal dropdown. That means that it is agnostic about what it is going to contain.
 
 It is intended to be a building block for more complex components but is perfectly usable. It is
--- a/ember-cli-build.js
+++ b/ember-cli-build.js
@@ -11,11 +11,7 @@ module.exports = function(defaults) {
     options.vendorFiles = { 'jquery.js': null };
   }
 
-  let app = new EmberAddon(defaults, {
-    babel: {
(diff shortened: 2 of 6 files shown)


cibernox:
@ksin Changes to the docs look good, but babel polyfill should not be required since the addon includes `ember-maybe-import-regenerator`. And tests always passed without the polyfill. How did you see that error.

ksin:
@cibernox I'm getting the error in local tests that use `await`. Pushing a commit up without the polyfill to see if it surfaces in travis

screen shot 2017-04-21 at 3 19 05 pm

cibernox:
As you see, tests pass in CI. I know what the problem is. It's a bug in ember-concurrency. Ignore it for now, it only happens when running the tests in development.

ksin:
Sounds good. Feel free to merge whenever. Could you share a little more details on the ember-concurrency bug? Just out of curiosity.

cibernox:
Yes. The version of ember-concurrency this addon was depending on used babel5. This addon uses babel 6 and the new [targets feature](http://rwjblue.com/2017/04/21/ember-cli-targets/) in ember-cli 2.13. 
For ergonomics, in development my targets are modern browsers that support async/await: https://github.com/cibernox/ember-basic-dropdown/blob/master/tests/dummy/config/targets.js#L9-L12 

Since `ember-maybe-import-regenerator` detects that for those targets the polyfill is not needed, it doesn't include it. But since ember-concurrency is not in babel6, it still transpiles code assuming that the polyfill is going to be used.

This doesn't happen in test environment because the targets include Internet exporer and the regenerator polyfill is added.

Updating to a version of e-c that uses babel6 should fix it.

cibernox:
Thanks! I've updated e-c in another PR.

PR #271, opened 2017-04-20
WIP: Positioning updates

RustyToms:
closes #270 

Adds the `targetContainerID` property to the basic-dropdown. This `targetContainerID` is used to change where the dropdown content is located using ember-wormhole.

The positioning logic is updated to account for the position and width of the new containing element.

PR #276, opened 2017-04-22
add scrolling container problem

RustyToms:
@cibernox here is a demo of the positioning and scrolling issues: http://localhost:4200/cookbook/scrolling-container

related to #271 and #268 

scrolling_demo


RustyToms:
I was moving commits between branches and I slightly messed it up. I just pushed up the proper commit

PR #277, opened 2017-04-22
Working example of scrolling containers

RustyToms:
Related to #276 

This PR is to demonstrate the proposed fixes at #268 and #271 
It combines the positioning fixes and scrollableElements listeners so that the demo at http://localhost:4200/cookbook/scrolling-container works.

working_scrolling_demo


cibernox:
Thanks. I'm going to play with this.

dossorio:
Hi @cibernox  !

Did you get to play with it ? we have got the same issue here, due to our layout.

Is the anything I can help with ?
Thanks !!

cibernox:
@dossorio This has been implemented. It was implemented in a slightly different fashion but I like the approach i followed better because users don't have to specify any option, the component automatically detects the scrollable parents.

I think we can close it.

xcambar:
@cibernox Where has it been implemented? I can't find it in the latest releases nor in master.

FWIW, this bug prevents us from upgrading from ember-power-select 1.7.x to 1.8.x.

cibernox:
@xcambar it has: https://www.ember-basic-dropdown.com/scrolling-container

PR #280, opened 2017-04-22
[BREAKING] Introduce `destination=`. Deprecate `to`.

cibernox:
Passing `to="id-of-elmnt"` to the `{{#dropdown.content}}` component is deprecated.
The way to specify a different destination for `ember-wormhole` is now by passing
`destination=id-of-elmnt` to the top-level `{{#basic-dropdown}}` component.
The old property `to=` still works but shows a deprecation message and will
be removed soon.

### DIFF
--- a/addon/components/basic-dropdown.js
+++ b/addon/components/basic-dropdown.js
@@ -5,7 +5,8 @@ import layout from '../templates/components/basic-dropdown';
 import { join } from 'ember-runloop';
 import fallbackIfUndefined from '../utils/computed-fallback-if-undefined';
 import calculatePosition from '../utils/calculate-position';
-const { guidFor } = Ember;
+import computed from 'ember-computed';
+const { guidFor, testing, getOwner } = Ember;
 
 const assign = Object.assign || function EmberAssign(original, ...args) {
   for (let i = 0; i < args.length; i++) {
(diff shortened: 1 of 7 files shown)


cibernox:
@RustyToms As part of work you've done it made clear that the ID of the element in which the dropdown is going to be appended must be passed to the top-level component instead of being passed to the `{{#dropdown.content}}` one, as it is needed to perform a correct positioning.

This PR makes that change and only that change, while allows the old method to work but shows a deprecation message.

addon/components/basic-dropdown.js
```diff
@@ -225,5 +236,13 @@ export default Component.extend({
       registerAPI(newState);
     }
     return newState;
+  },
+
+  _getDestinationId() {
+    if (testing) {
+      return 'ember-testing';
```
RustyToms: While testing you would expect the dropdown to work the same as you would in production. And the client app can make sure that the element they want the dropdown appended to is available in the test environment. But I understand you probably don't want to render the dropdown on the `body` during testing. I'd recommend that *if* `destination` is provided, it be used regardless of whether the app is in testing or not. If `destination` is not provided, then change it to `ember-testing`.

However, if you think that it is more important to make testing simpler, it might be better to just set `renderInPlace` true while testing.

PR #281, opened 2017-04-23
[BREAKING] Account for the containers X/Y position when calculating c…

cibernox:
…oordinates

The `calculatePosition` function now checks if the parent of the destination div has
position absolute or relative, and if so, substracts its X/Y coordinates to correct
the positioning of the content.
The breaking part is that now the signature of the function is:

`calculatePosition(trigger, content, destination, options)` instead of
`calculatePosition(trigger, content, options)`.

Closes #233


cibernox:
@RustyToms This old bug was somewhat related to your goal, so I had to fix it first.

addon/utils/calculate-position.js
```diff
@@ -24,14 +25,22 @@ export default function(_, _2, { renderInPlace }) {
...
+export function calculateWormholedPosition(trigger, content, destination, { horizontalPosition, verticalPosition, matchTriggerWidth, previousHorizontalPosition, previousVerticalPosition }) {
   // Collect information about all the involved DOM elements
   let scroll = { left: window.pageXOffset, top: window.pageYOffset };
   let { left: triggerLeft, top: triggerTop, width: triggerWidth, height: triggerHeight } = trigger.getBoundingClientRect();
   let { height: dropdownHeight, width: dropdownWidth } = content.getBoundingClientRect();
   let viewportWidth = self.document.body.clientWidth || self.window.innerWidth;
   let style = {};
 
+  // Apply containers' offset
+  let destinationParentPosition = window.getComputedStyle(destination.parentNode).position;
```
RustyToms: Maybe a `getPositionParent` could be called? So that the dev doesn't have to worry about which node they are adding the dropdown to and what the styling is, but the dropdown will automatically position itself correctly no matter where it is.

PR #282, opened 2017-04-23
[PERF] Optimize number of reposition calls

cibernox:
If the wormhole destination is a direct child of the BODY element, there
is no need to reposition it after every scroll event, since the origin
of coordinates is the scrollable element itself.

This is the case for almost every dropdown. That method, albeit fast, can be called tens of times per second, so I expect this to matter quite a lot.


cibernox:
@RustyToms Also somewhat related to the reposition logic. I've discovered that on the most common use-case the addon was doing unnecessary work.

PR #283, opened 2017-04-23
Attempt automatic nested scroll detection (N levels)

cibernox:
@RustyToms I think I've squared the circle finally. 

This approach in your demo seems to work automatically, without forcing the user to pass **any** option. Not even `destinatination`. I did it by automatically find all the parents with scroll and reposition when any of them scrolls.

Can you give it a try?

### DIFF
--- a/addon/components/basic-dropdown.js
+++ b/addon/components/basic-dropdown.js
@@ -90,20 +90,12 @@ export default Component.extend({
   },
 
   // CPs
-  defaultDestination: computed(function() {
-    if (testing) {
-      return 'ember-testing';
-    }
-    let config = getOwner(this).resolveRegistration('config:environment');
-    return config['ember-basic-dropdown'] && config['ember-basic-dropdown'].destination || 'ember-basic-dropdown-wormhole';
-  }),
-
   destination: computed({
     get() {
-      return this.get('defaultDestination');
+      return this._getDestinationId();
     },
     set(_, v) {
-      return v === undefined ? this.get('defaultDestination') : v;
+      return v === undefined ? this._getDestinationId() : v;
     }
   }),
 
@@ -179,10 +171,10 @@ export default Component.extend({
       return;
     }
 
-    let destination = self.document.getElementById(this.get('destination'));
+    this.destinationElement = this.destinationElement || self.document.getElementById(this.get('destination'));
     let options = this.getProperties('horizontalPosition', 'verticalPosition', 'matchTriggerWidth', 'previousHorizontalPosition', 'previousVerticalPosition', 'renderInPlace');
     options.dropdown = this;
-    let positionData = this.get('calculatePosition')(triggerElement, dropdownElement, destination, options);
+    let positionData = this.get('calculatePosition')(triggerElement, dropdownElement, this.destinationElement, options);
     return this.applyReposition(triggerElement, dropdownElement, positionData);
   },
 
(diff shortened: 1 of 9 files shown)


cibernox:
I'm going to release this as a beta so people can check that it doesn't break.

RustyToms:
Ok I'll check it out and try it in our app as soon as I can

RustyToms:
A couple of things:
The positioning is still broken if you supply a `destination` that is not the same size, shape, and position of the `body`.


For the scrolling, this only works if you *don't* supply a `destination` if you add a destination. With this approach you need to choose if you are going to use a `destination` or deal with a scroll, you can't do both. That would be fine (assuming it is added to the docs), and is easiest to implement, except for the performance hit of always listening to all scrolls between the dropdown and `body`. Most of the time if there is scrolling, you'd want to use `destination` to put the dropdown inside whatever is scrolling, but now if you do that the positioning gets messed up.
with_destination
Basically, it breaks `destination` on any page that scrolls.
I'd either get rid of `destination` all together, or refactor the scrolling/positioning so that they work together.

If you don't want the dev to have to identify the scrolling elements, you would need to find the closest common ancestor between the `destination` and the trigger. And find all scrolling elements between the trigger and the common ancestor, and also between the `destination` and the common ancestor.

cibernox:
I'm sitting on the fence here. I think that this approach works and is a no brainer for 99% of the people. 

Actually the original intent of `destination` was not to allow embedding the dropdown in scrollable elements, but allow the users whose applications are not direct child of the body, or even pages with several applications (using ember-islands p.e) to specify a common container as close to the root of the page as possible.

I think that the less people have to worry about destination, the better. I wouldn't remove it entirely tho, it still has its utility.

For the rest of it, did this work as intended without specifying any target? I want to make the component as unattended as possible, because every option added to this component also means another option added to `ember-power-select`. The less options the better.

RustyToms:
Yea, if the dropdown is on the `body` it works. Keep in mind the point I am making is that if anything scrolls, including the `body`, if you are using `destination` or `to` it breaks, so it isn't just an inconvenience. I would say this dropdown would only work if the dropdown is on the `body`. I think for my team we would consider still using my fork instead of the main repo here because of the performance implications. I'm not sure, I'd much rather use the main repo.

I understand your dilemma here. I have a lot of experience dealing with this sort of thing, but not with creating and maintaining an addon. If you tell me what kind of API you want, what kind of coding style you want, and the end result you want to see I can make a PR that will do everything you want. I don't think there is anything technically that is keeping this from happening. I think everything could be accomplished with the same API (just a new `destination` property) that exists in this PR.

But if you aren't interested I won't bother you or spend the dev time on it. Let me know what you think.

RustyToms:
@cibernox by the way, I didn't make it clear, I appreciate the time you are putting into this. No matter what this has been a big help to me and my team, as I am sure it has been to many others. You are very responsive and have been working at this quickly, thanks a lot!

cibernox:
It would only break if you put the destination inside a scrollable element. 
My question is: is there any good reason to put the dropdown inside an scrollable element if placing it in the body works out of the box, no matter in how how many nested scrollable elements your dropdown is placed in?

I struggle to find a good reason to do so.

Performance should be the same. It is true that we add more event listeners, but those events listeners are only added while a dropdown is opened, and given the nature of scrollable elements, only one scrolls at the same time, so you will not fire scroll events at a higher rate than you usually do.

RustyToms:
@cibernox yes, I understand that, I think it is a valid approach. I think that is how select2 works, I think it is always on the `body` tag. But you still need to fix the `destination` positioning, it has the same problem as before. If the element it is going to isn't the same size, shape, location of the `body` tag the dropdown isn't positioned correctly. Or maybe just remove `destination` entirely if you don't want to support that. https://github.com/cibernox/ember-basic-dropdown/pull/284 Let me know if you want me to make any code for that.

PR #285, opened 2017-04-25
[Docs]: Update README to match API

alexander-alvarez:
to match
https://github.com/cibernox/ember-basic-dropdown/blob/master/addon/utils/calculate-position.js#L107


cibernox:
Thanks!

PR #288, opened 2017-04-26
Fix postion calculation - ensure that tagName are always upperCase

cbou:


cbou:
In my app (with last Chrome) `document.body.tagName` returns "body" and not "BODY". I guess it's because of the DTD or something like that...

cibernox:
@cbou What does DTD stands for? In the html spec `tagName` is always uppercase.

cbou:
Please try on this page: https://charlesbourasseau.com/index.xhtml

`document.body.tagName` output `body` and not `BODY`

cibernox:
@cbou I see. that is because your page is not HTML but XHTML, so it doesn't follow the HTML spec.

I'll merge this.

cbou:
Yes it has nothing to do with DTD actually ;)

Thanks

PR #292, opened 2017-05-04
Update native-dom-helpers to 0.4.0

cibernox:


cibernox:
@RuslanZavacky This should fix your problem BTW.

I asked and basically, there is no way that is not **incredibly** hacky of forcing one version of a dependency to win over another. The right way in case of a discrepancy is to specify the version in your app, and that will win. 

However I was pointed to https://github.com/salsify/ember-cli-dependency-lint which can help you detect when you have incompatible dependencies. It will just warn you and its up to you to make a PR to the addon that is pined on an old version of a dependency.

RuslanZavacky:
Thank you @cibernox :)

PR #299, opened 2017-06-01
Move ember-native-dom-helpers to devDeps

simonihmig:
This included them into an app's `test-support.js` although the app does not use/need them...


cibernox:
Those helpers are used from within `ember-basic-dropdown`'s helpers. If this is a devDependency that would fail, right?

simonihmig:
You are right, did a quick check and thought there were just used internally, but missed this (https://github.com/cibernox/ember-basic-dropdown/blob/3314838916f170459417d3218795f2d3e9bd7acc/test-support/helpers/ember-basic-dropdown.js#L3).

PR #302, opened 2017-06-05
Update deps to fix test suite

cibernox:


cibernox:
@sukima I've seen that you had the same error How did you fix it. I tried the `--pure-lockfile` and removed `yarn: true` from the cache. Any other idea?

sukima:
@cibernox I had to,

1.  Remove the `cache:` node from `.travis.yml`
2.  Add `--pure-lockfile` to `.travis.yml` during the yarn install
3.  Go into Travis console and *delete* all the caches

It took a few tries and I'm still not sure if it clear itself or my melding had something to do with it. Also if worse comes to worse may be `yarn install --force --pure-lockfile` would help.

sukima:
This is yet another coffin why I _loathe_ Sass and its insistence to compile machine code for an interpretative language that could easily be implemented in the native language of the addon/gem. TL;DR Sass continues to be problems; its failure/success rate seems disproportional.

cibernox:
@sukima I wiped caches, and I think I'm following all your steps. Do you see anything I'm missing?

sukima:
I don't see anything that stands out. Is it possible ember-try is not yarn compatible?

PR #325, opened 2017-07-10
Fix mobile Firefox issue with invoking the click event

jdurand:
Investigated because I was receiving tons of `Not enough arguments to MouseEvent.initMouseEvent` in production from mobile Android Firefox browsers and stumbled on this: 
https://stackoverflow.com/questions/26761625/not-enough-arguments-to-mouseevent-initmouseevent?answertab=active#tab-top


cibernox:
The problem is that if you fix this, you break internet explorer. And firefox mobile is testimonial. What about a try/catch with both approaches?

jdurand:
Fair enough 👍

cibernox:
Thanks!

PR #333, opened 2017-07-17
Fixes the dropdown being incorrectly positioned when the body is position:: relative

### LINKED ISSUE #332: position: relative on the body element causes badly positioned element
[Minimal repro twiddle](https://ember-twiddle.com/f594425bff4c5d77a7b5abf24cc88727?openFiles=templates.application.hbs%2C)

When the body element has `position: relative`, ember-basic-dropdown has issues with calculating the scroll.

The code causing this is [here](https://github.com/cibernox/ember-basic-dropdown/blob/master/addon/utils/calculate-position.js#L108). Removing `scroll.top` fixes it (but only when you've styled the `body` elem like this.

NLincoln:
Fixes #332

### DIFF
--- a/addon/utils/calculate-position.js
+++ b/addon/utils/calculate-position.js
@@ -105,7 +105,17 @@ export function calculateWormholedPosition(trigger, content, destination, { hori
   }
 
   // Calculate vertical position
-  let triggerTopWithScroll = triggerTop + scroll.top;
+  let triggerTopWithScroll = triggerTop;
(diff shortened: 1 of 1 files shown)


cibernox:
I'm in the middle of a moving so I'll have 0 time until next week. Please ping me again if I haven't responded by next Wednesday.

cibernox:
Checked locally. Nice fix!

PR #334, opened 2017-07-19
fixed a bug when el is document and doesn't have classList

jinkea:
this fixed:
when dropdown is open and clicking on the scrollbar on page, a JS error is generated:
screen shot 2017-07-19 at 10 22 54 am


cibernox:
@jinkea So, the scroll bar is some sort of special HTMLElement without classList? Really?

That is super weird, can you confirm?

jinkea:
@cibernox el is just document when you click on scrollbar, Chrome on Mac

cibernox:
TIL. Thanks!

PR #343, opened 2017-09-11
Add css fix from previous commit into less file

g-cassie:
There was a fix for a css bug in efc059a047085065624ec281dbbb9891b885e325 that never ended up in the `.less` file.  This adds it.


cibernox:
Thanks! Sorry for the delay, I just forgot about this.

PR #351, opened 2017-10-23
Prevent scrolling as an option.

nathanhammond:
Platform-supplied `select` boxes do not allow scrolling once a `select` is open. This approximates that behavior. (Doesn't work beyond a frame boundary, but works completely within the present `window` context.)

### DIFF
--- a/addon/components/basic-dropdown/content.js
+++ b/addon/components/basic-dropdown/content.js
@@ -263,8 +263,10 @@ export default Component.extend({
   wheelHandler(event) {
     const element = this.dropdownElement;
     if (element.contains(event.target) || element === event.target) {
+      // Discover the amount of scrollable canvas that is within the dropdown.
       const availableScroll = getAvailableScroll(event.target, element);
 
+      // Calculate what the event's desired change to that scrollable canvas is.
       let deltaX, deltaY;
       if (event.deltaMode === 0) {
         // DOM_DELTA_PIXEL: applies almost everywhere.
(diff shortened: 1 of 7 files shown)


cibernox:
I pulled the changes and I found a bug. Since you are allowing scroll events when the event's source is within the content, once the scroll within the content reaches the bottom or the top, the entire page scrolls next.

I believe that you need to get the scroll position of the closest scrollable div from the event's source. If it's at the bottom of its scroll, you have to prevent scroll events with a negative Y. If it's at the top, you have to prevent scrolls with a positive Y.

cibernox:
I finally understood the code and it seems reasonable. I only see a possible solution that could be less complicated, if you tried it and it didn't work I'll merge it.

There is an option already on this component to add an invisible (unless you style it) overlay that covers the entire viewport. Would it be possible to simply capture and prevent the bubbling of wheel/touchmove events on this overlay? If so, I'd be confortable with forcing the invisible overlay.

nathanhammond:
Your proposal is functionally identical to the approach that [I initially tried](https://github.com/cibernox/ember-basic-dropdown/pull/351/commits/7e4f05e1a31a3af3e605ecff35981d1e95e32de7) and suffers from the bug you identified which pushed me down this path to begin with.

Unfortunately I believe that this is the only viable solution unless we get [`document.rootScroller`](https://www.chromestatus.com/feature/4533350404456448) ... but even still we likely wouldn't be able to rely on that until 4 years from now.

nathanhammond:
(Note: I dropped `touchmove` event handling because we're already different from native select behavior in touch-based browsing contexts and I didn't want to constrain panning.)

cibernox:
In!

PR #356, opened 2017-11-06
Position calculation for `right` and `auto-right` when renderInPlace=true 

### LINKED ISSUE #255: horizontalPosition is ignored when you renderInPlace
setting `renderInPlace=true` and `horizontalPosition='right'` doesn't seem to put the content right-aligned but if `renderInPlace` is false then it is correctly aligned.

graemechapman:
Closes #255

### DIFF
--- a/addon/utils/calculate-position.js
+++ b/addon/utils/calculate-position.js
@@ -152,7 +152,14 @@ export function calculateInPlacePosition(trigger, content, destination, { horizo
     let { width: triggerWidth } = trigger.getBoundingClientRect();
     let { width: dropdownWidth } = content.getBoundingClientRect();
     positionData.style = { left: (triggerWidth - dropdownWidth) / 2 };
+  } else if (horizontalPosition === 'auto-right') {
+    let triggerRect = trigger.getBoundingClientRect();
+    let dropdownRect = content.getBoundingClientRect();
+    positionData.horizontalPosition = triggerRect.right > dropdownRect.width ? 'right' : 'left';
+  } else if (horizontalPosition === 'right') {
+    positionData.horizontalPosition = 'right';
   }
+
   if (verticalPosition === 'above') {
     positionData.verticalPosition = verticalPosition;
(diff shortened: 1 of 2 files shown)


cibernox:
Sorry for delaying this so much. I'm going to test this now 👍

cibernox:
@graemechapman can you rebase? I've improved the demo for the positioning to allow the user to toggle between in-place/wormhole. That will make the testing much easier.

graemechapman:
@cibernox rebased :slightly_smiling_face:

cibernox:
If you go to http://localhost:4200/docs/position you can play with all the combinations of positions, with and without `renderInPlace`. The behaviour with renderInPlace seems still broken.

graemechapman:
> If you go to http://localhost:4200/docs/position you can play with all the combinations of positions, with and without renderInPlace. The behaviour with renderInPlace seems still broken.

Ah... I'll check it out, thanks!

graemechapman:
@cibernox it's adding the right class names - I think the issue might be more to do with the css in the demo page? The in place dropdown gets a position of absolute, but there is no position on the parent so it is aligning to the body. Adding 
```css
    position: relative;
    display: inline-block;
```
to the `button-pull-left` element works - happy for me to include this?

cibernox:
Ideally this should work automatically. It's hard to document that wherever they use this positioning with `renderInPlace=true` they must add certain classes to the parent div.
Perhaps the component could have a wrapper around both the trigger and the content when it's rendered in-place?

cibernox:
After some experimentation I found that the demo was slightly broken because `renderInPlace` is not a property you're supposed to mutate, so the markup being rendered wasn't realistic.

Also, I think this PR is correct. This approach works in all cases except when the trigger is smaller than the full-width of the container on purpose. Since the user has to do that on purpose, I think it's also fair to force them to constraint the size of the wrapper on purpose.

graemechapman:
Thanks @cibernox, yes I see that `renderInPlace` is referenced on init, makes sense that it needs a `tagName` for this to work 🙂

PR #357, opened 2017-11-06
Have basic-dropdown trigger aria-role default to 'button'

brandynbennett:
This will make it possible for parent components to pass in a value for
this and they won't have to worry about it breaking if the value is not
defined


cibernox:
Nice. Can you only add a test in https://github.com/cibernox/ember-basic-dropdown/blob/master/tests/integration/components/basic-dropdown/trigger-test.js ?

PR #367, opened 2017-12-15
Remove Support for non-glimmer versions of Ember (<= 2.9)

cibernox:


cibernox:
@simonihmig I believe this will fix the problem with changing the root element for testing

cibernox:
Ok. It's time to break with the past. Ember 2.0, released mid October 2016, won't be compatible. Ember 2.10, the first release including Glimmer2, released in November 2016 will continue to work.

One year seems enough support, and this allows this addon to drop ember-wormhole and use the built-in `#-in-element` keyword in glimmer2. This saves a few Kb in the app and its also more efficient than wormhole.

This is a breaking change, so I'll kick in the 1.0.0 cycle with some beta version.

PR #377, opened 2018-01-15
Allow string values for positional CSS attributes.

Serabe:
When dealing with complex `calculatePosition` functions,
we may need to use non-numerical values for `top` / `bottom` /
`right` / `left` to overwrite previous values. An example
would be a case where we would have the content vertically
centered to the right side of the trigger (custom numerical
values) but if it does not fit, we would have it `right` and
`below` (no custom values).

The way we can fix this on the custom function is to return
`auto` values for such properties when not needed.

Ember Basic Dropdown would take these values an append `px`
at the end. While this still works, warning were annoying
and still it is not valid for different values.


Serabe:
Closing this.

PR #383, opened 2018-02-23
move ember-native-dom-helpers to devDependencies

akatov:


cibernox:
I believe that we can't remove it yet, because it's used in this file: https://github.com/cibernox/ember-basic-dropdown/blob/master/addon-test-support/helpers.js#L4

I'd like to use `import { click } from '@ember/test-helpers';` but probably a lot of people haven't updated. Since including it doesn't ship any code to production I don't think it's a big deal to install one extra dependency. But I do want to remove it in a few weeks.

PR #394, opened 2018-04-05
Allow arbitrary CSS rules when repositioning

cibernox:


raido:
There seems to be left over console.log that fails lint on Travis.

PR #395, opened 2018-04-08
Use click from `ember-native-dom-helpers` if Ember Version < 3.0

esbanarango:
Similar to this issue: https://github.com/cibernox/ember-power-select/issues/1085

If Ember version is < 3.0, the `click` from `@ember/test-helpers` throws `Must setup rendering context` error.

This PR lets projects using ember versions < 3.0 upgrade to the new Ember's latest testing APIs, based on RFCs [232](https://github.com/emberjs/rfcs/blob/master/text/0232-simplify-qunit-testing-api.md) and [268](https://github.com/emberjs/rfcs/blob/master/text/0268-acceptance-testing-refactor.md).


blimmer:
👍

PR #408, opened 2018-05-14
allows CSS selector to define destination element

### LINKED ISSUE #406: define destination using CSS selector instead of element Id
I need my dropdowns appended directly to the `<body>` instead of inside some "wormhole" element (in that case there were CSS conflicts that misaligned the dropdowns).

In order to achieve this I had to override some ember-basic-dropdown Js so that the user can pass a CSS selector as a _destination_ instead of a unique Id.

```Javascript
import EmberBasicDropdown from 'ember-basic-dropdown/components/basic-dropdown';
import { computed } from '@ember/object';
import { getOwner } from '@ember/application';

export default EmberBasicDropdown.extend({
  _getDestinationSelector() {
    let config = getOwner(this).resolveRegistration('config:environment');
    return config['ember-basic-dropdown'] && config['ember-basic-dropdown'].destination || 'body';
  },
  destination: computed({
    get() {
      return this._getDestinationSelector();
    },
    set(_, v) {
      return v === undefined ? this._getDestinationSelector() : v;
    }
  }),
  reposition() {
    this.destinationElement = document.querySelector(this.get('destination'));
    this._super(...arguments);
  }
});
```

and

```Javascript
import EmberBasicDropdownContent from 'ember-basic-dropdown/components/basic-dropdown/content';
import { computed } from '@ember/object';

export default EmberBasicDropdownContent.extend({
  destinationElement: computed('destination', function() {
    return document.querySelector(this.get('destination'));
  })
});
```

is this a safe override and if so, could I make a PR to turn the above into an optional feature of this addon?

Pixelik:
Closes #406 

1. the user can continue to use `destination` in order to define the id of the destination element
2. the user can now also use `destinationSelector` in order define a destination element via a CSS selector

### DIFF
--- a/addon/components/basic-dropdown.js
+++ b/addon/components/basic-dropdown.js
@@ -102,13 +102,13 @@ export default Component.extend({
   },
 
   // CPs
-  destination: computed({
-    get() {
-      return this._getDestinationId();
-    },
-    set(_, v) {
-      return v === undefined ? this._getDestinationId() : v;
+  computedDestinationSelector: computed('destination', 'destinationSelector', function() {
+    if (this.get('destination')) {
+      return `#${this.get('destination')}`;
     }
+    return this.get('destinationSelector') ||
+           this._getDestinationSelectorFromConfig() ||
+           '#ember-basic-dropdown-wormhole';
   }),
 
   // Actions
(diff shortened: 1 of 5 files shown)


cibernox:
Is there a chance you can use still `destination`, but detect that if the first character of the destination is not a letter (which would make it an invalid element ID) it assumes it's a CSS selector?

Pixelik:
I thought about it but the problem is we have no way of distinguishing between a `destination` that is supposed to be an id and a `destination` that is supposed to be a CSS selector targeting an element by its tag name, e.g. `<div id="hello"></div>` and `<hello></hello>`. Granted a rare case, but possible nonetheless.

Enjoy your holidays 😎

Pixelik:
Hello, any chance of this PR getting merged ? :)

cibernox:
@Pixelik I forgot about this. I believe that targeting a tagName is a rather edge case, and not allowing users to pass IDs could be a breaking change, so I'd still use the first character to decide wether the value is an `id` or a css selector.

Pixelik:
to be honest, the whole reason behind this PR is that we want to be able to target the <body> as a destination for a modal and the  css selector for body does not starr with any special characters.. :-/

cibernox:
That is actually somewhat dangerous, since `{{in-element destination}}` doesn't guarantee that the content inside the destination element is not wiped entirely before insertion. Right not because of implementation details it happens to be respected, but I've been warning that it might change, as it's undefined behavior.

Using a dedicated target element is safe and there is no content inside that may be wiped.

Issue #409, opened 2018-05-15
Using the browser scrollbar closes the dropdown


blagh:
Although you can still scroll the page with a mouse scrollwheel for example, it's a bit of a usability issue if you can't also use the scrollbars built into the browser.

I've added a gif to illustrate -- this is more of a problem with bigger dropdowns.

<img src="https://user-images.githubusercontent.com/5570885/40081847-328252b8-585d-11e8-8570-91864f4b7e24.gif" width=200px>

cibernox:
I think is unavoidable really as clicking outside the component closes it. You can make the component stay open by returning `false` from the `onClose` action (see http://ember-basic-dropdown.com/docs/dropdown-events). 

That action will receive the mousedown event as second argument, but _i don't think there is a way to detect the user has clicked in a scroll bar_. If I'm wrong please tell me.
I don't know if it's possible to detect that the user clicked

cibernox:
I see. The reason why this component uses onmousedown is because it was _initially_ extracted from Ember Power Select, and selects open and close on mousedown, not on click. Since this component is more generic than a select opening on click should, at least, be an allowed option which would also fix your complaint.

I'm on holidays right now, but it seems like a feature request.

blagh:
By all means, have a good holiday! Thanks for checking in, in the meanwhile :grin:

PR #427, opened 2018-08-29
Add config option to fix scrollbar clicking

k-fish:
As mentioned in #409, clicking a scrollbar will close an open basic-dropdown. This is due to mousedown being used instead of click. To not break existing behaviour, I've added a config option which can be passed to the parent container (basic-dropdown) or the content component.

If you think this looks reasonable I'll go ahead and add notes to the docs as well.


wcarss:
:+1: this would be super helpful, @cibernox do you have any thoughts? It seems like it would address https://github.com/cibernox/ember-basic-dropdown/issues/409

cibernox:
Sorry, I missed this. Twice.
I agree with the PR, I just want to suggest a small naming change. I'd change the `useClickEvent` to something something more specific like `rootEventType`, that defaults to `mousedown` but can be changed.

Also, I think this behavior seems a better default, so please leave a `TODO` in the code saying something `TODO: Change default event type to "click" in 2.0`

cibernox:
Could you check tests?

tests/integration/components/basic-dropdown/content-test.js
```diff
@@ -103,12 +103,54 @@ module('Integration | Component | basic-dropdown/content', function(hooks) {
     await render(hbs`
       <div id="destination-el"></div>
       <div id="other-div"></div>
-      {{#basic-dropdown/content dropdown=dropdown destination='destination-el'}}Lorem ipsum{{/basic-dropdown/content}}
+      {{#basic-dropdown/content rootEventType='mousedown' dropdown=dropdown destination='destination-el'}}Lorem ipsum{{/basic-dropdown/content}}
```
k-fish: @cibernox I added the `rootEventType` here because the default is defined in the `basic-dropdown` component. Is this the correct pattern or should the fallback be moved down into the content? Content shouldn't really exist without the parent component binding, so I figure this is ok, but I'm not sure whether to add this for the remaining tests to ensure click behaviour isn't unexpected.

cibernox:
Implemented in #438

cibernox:
Published in 1.1.1

PR #446, opened 2019-01-25
ROAD TO 2.0 - Convert to Octane

cibernox:


gossi:
what about using pods? = more prepared for octane 😁

cibernox:
@gossi in an addon with three classes I don't see the point. When the MU is working (is it already? Last time I checked addons didn't work) I'll just run the codemod.

gossi:
No, it is just standard ember, you can have `component.js` and `template.hbs` in the same folder, like this:

```
addon/
  components/
    basic-dropcdown/
      component.js
      template.hbs
```

PR #452, opened 2019-03-08
Adds "https" to ember-twiddle link in readme

ryanlabouve:
Without this prefix, the link get's built to `<a href="/cibernox/ember-basic-dropdown/blob/master/www.ember-twiddle.com">Ember Twiddle</a>` instead of actually linking to twiddle.


cibernox:
Thanks!

PR #461, opened 2019-06-06
Fix undefined events

wukongrita:
Fix undefined events when is setted with rootEventType


mkszepp:
very old PR, but we handle the undefined already

Issue #431, opened 2018-10-12
EmberError - You must use set() to set the `previousHorizontalPosition` property


jaredgalanis:
I'm getting the following error when attempting to click close on the target. Is this a known issue?

`"Assertion Failed: You must use set() to set the `previousHorizontalPosition` property (of <my-app@component:basic-dropdown::ember519>) to `null`."`

jaredgalanis:
Closing because I'm no longer experiencing this issue. Not 100% sure what it was, but I'll comment further if I find out.

BnitoBzh:
I have the same issue with Ember 3.10.2.
```
  <BasicDropdown as |dd|>
    <dd.trigger>{{t (concat "language." this.intl.locale.firstObject)}}</dd.trigger>
    <dd.content>
      {{#each this.languages as |language|}}
        <div {{action "setLanguage" language}}>{{t (concat "language." language)}}</div>
      {{/each}}
    </dd.content>
  </BasicDropdown>
```

cibernox:
@BnitoBzh version of the addon?

BnitoBzh:
@cibernox EBD@1.1.2

BnitoBzh:
@cibernox i want to submit a pull-request for EBD@1.1.2 but there is no active branch for this version .. The master branch is on @2.x ...
On which branch can i made my pull request ?

cibernox:
@BnitoBzh I created a 1.X branch. Try with that branch and I'll try to port the change to 2.X

PR #463, opened 2019-06-24
Fix set issue #431

BnitoBzh:
issue #431


cibernox:
If you know the reproduction steps, can you include a failing test that passes with your changes?

BnitoBzh:
i have no reproduction step ... i use these addon in 3 applications and i have this error in only one ...

cibernox:
Released 1.1.3

PR #475, opened 2019-07-08
Rename `onOpen` to `willOpen`

cibernox:


cibernox:
After some thought, and given that there is no need for a `didOpen` and `didClose` (that behavior can be attained with `{{did-insert}}`/`{{will-destroy}}` on the `<dd.Content>` component, I don't think that it's worth making a breaking change to clarify that those methods run before the component is opened.

Issue #478, opened 2019-07-13
Expose calculate-position util as public API?


Kilowhisky:
Thanks for making this plugin, its been great working with it! I just upgraded to 2.0.0 and i like it!

Recently i have gained a need to make custom positioning of the input on mobile clients. (basically dropdowns on mobile need to be a modal like popup in the center of the screen in order to facilitate better usability).

Anyways i see that the API has a calculateModalPosition function that is perfect for override. Great! 

But, i only want to override it under certain scenarios. Anything that isn't mobile i want to keep the old behavior.  

I see that the override is called like this.

```
let positionData = (this.calculatePosition || calculatePosition)(triggerElement, dropdownElement, this.destinationElement, options);
```

Can the `calculate-position` util be part of the public API so i can call it when i'm done with my override work?

So basically i'd like to define my method like this. (i'm overriding because it needs to be app wide and i do some other 'common' initialization).

```
import BasicDropdown from 'ember-basic-dropdown/components/basic-dropdown';
import calculatePosition from 'ember-basic-dropdown/utils/calculate-position';

export default BasicDropdown.extend({
    calculatePosition(trigger, content, destination, args) {
        if(window.width > 800 || !mobileModal) {
            return calculatePosition.apply(...arguments);
        }
        // do custom logic
    }
});
```

cibernox:
Totally. In fact I think it is an oversight, I intended to make it public. Feel free to create a PR exporting it in the /app folder

PR #479, opened 2019-07-14
export calculate-position as public api

Kilowhisky:
Refs #478


cibernox:
Thanks! I just realized, could you create another PR adding a test that uses the util (no need to assert the correctness of the resukt) just to ensure we don't remove this public API

cibernox:
Thanks! I just realized, could you create another PR adding a test that uses the util (no need to assert the correctness of the resukt) just to ensure we don't remove this public API

PR #480, opened 2019-07-14
added test for existance of calculate position

Kilowhisky:
Refs #478  and #479 

I took a stab at it. Not sure how to qualify it as 'public' API.
sorry, don't have much experience with tests.


cibernox:
That is exactly that I was looking for! Thanks

cibernox:
@Kilowhisky I've released 2.0.1 with this new public API now.

Issue #420, opened 2018-08-04
Possibility of splitting onOpen & onClose to before & after actions?


sdhull:
First I want to say thank you for this most excellent ember addon, which is so foundational to many other addons!!

Second I wanted to ask if you'd be open to deprecating `onOpen` and `onClose` in favor of `beforeOpen` and `beforeClose` as well as adding `afterOpen` and `afterClose` hooks.

Or if it's possible to achieve `after` hooks some other way I'd be very interested to learn!

sdhull:
In case it wasn't clear -- if you're open to considering a PR that would do this, I'm happy to take a stab at implementing it and opening a PR

JenLyndle:
the afterOpen and afterClose hooks will be very helpful!

hadiwina:
I agree with @sdhull's suggestion. The split up actions are gonna be super helpful!

Thanks for the great addon! :)

cibernox:
I think it's a good idea, but I'm short of time lately.

The approach would be to add a `onBeforeOpen`/`onAfterOpen` and `onBeforeClose`/`onAfterOpen`, and, for backwards-compatibility, map the current `onOpen`/`onClose` to `onBeforeOpen`/`onBeforeClose` with a deprecation, to eventually be removed in 2.0.

If anyone wants to take a stab let me know, I might be able to assist

sdhull:
I will likely take a stab at it this week.

PR #494, opened 2019-09-17
Issue #420 - Splitting onOpen and onClose to before and after actions

Ryan-Bell:
Issue #420 

My attempt at implementing onBeforeOpen, onBeforeClose, onAfterOpen, onAfterClose hooks.

I wasn't sure what arguments should be passed to the "after" hooks and whether or not `onAfterClose` should be called if the dropdown has been destroyed.

Let me know if there are any changes to documentation/implementation or tests you want.

Thanks for the great addon!

### DIFF
--- a/addon/components/basic-dropdown.js
+++ b/addon/components/basic-dropdown.js
@@ -10,6 +10,7 @@ import templateLayout from '../templates/components/basic-dropdown';
 import calculatePosition from '../utils/calculate-position';
 import { assign } from '@ember/polyfills';
 import requirejs from 'require';
+import { deprecate } from '@ember/application/deprecations';
 
 const ignoredStyleAttrs = [
   'top',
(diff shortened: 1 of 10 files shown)


addon/components/basic-dropdown.js
```diff
@@ -89,10 +90,23 @@ export default @layout(templateLayout) @tagName('') class BasicDropdown extends
...
+      if (this.onOpen(this.publicAPI, e) === false) {
+        return;
+      }
+    }
+    if (this.onBeforeOpen && this.onBeforeOpen(this.publicAPI, e) === false) {
       return;
     }
     this.updateState({ isOpen: true });
+    if (this.onAfterOpen) {
+      this.onAfterOpen(this.publicAPI, e);
```
cibernox: I'm not 100% if we should call it here or wait until the `afterRender` step. 
I lean towards your solution, and let the user wait for the render if they want to.

cibernox:
This looks great. I just have one request. Can you update Ember Power Select to use the new onBeforeOpen/onBeforeClose hooks? Otherwise there will be a lot of noise. After that I'll merge it

cibernox:
It's also worth noting that I'm planing a new major release updating to GlimmerComponents, and it would be a good moment to remove the now-deprecated onOpen/onClose

Ryan-Bell:
Awesome! Quick question, should onAfterOpen and onAfterClose be exposed in ember-power-select or is it enough to update the onOpen and onClose internally? I noticed `power-select-multiple.hbs` has `@onClose={{@onClose}}`. Will this have to be updated to `@onBeforeClose={{@onClose}}`

hadiwina:
Is this PR no longer be merged?

PR #499, opened 2019-10-22
498 add event handlers

### LINKED ISSUE #498: Add back eventHandler arguments in V2
V2 removed the various event handler arguments (`onMouseEnter`, `onFocusOut` etc) in favour of the `{{on ...}}` element modifier.

This was to better align the add-on with Octane patterns.

However, if an end-user extends `ember-basic-dropdown` to add extra default behaviour for some internal dropdown component, there is a problem in so far as that element modifiers are not usable with the `(component)` helper and the extending component must yield out the `Content` & `Trigger` components, in the same was as `ember-basic-dropdown` does: https://github.com/cibernox/ember-basic-dropdown/blob/master/addon/templates/components/basic-dropdown.hbs

The problem is further excaberated by attempting to upgrade to Ember 3.13, where `ember-basic-dropdown` is required because of various deprecations in Ember.

This particular issue seems to be one that is being addressed via an RFC (https://github.com/emberjs/rfcs/issues/497), but until then, this leaves end users such as myself stuck not being able to upgrade to Ember 3.13.

As such, I'd like to propose adding back the arguments for the various event handlers, which can be passed through to various `{{on ...}}` modifiers on `Content` & `Trigger, something like: 

`templates/components/basic-dropdown-content.hbs`
```hbs
<Element
  id={{this.dropdownId}}
  class="ember-basic-dropdown-content ember-basic-dropdown-content--{{@hPosition}} ember-basic-dropdown-content--{{@vPosition}} {{this.animationClass}}{{if @renderInPlace " ember-basic-dropdown-content--in-place"}} {{@defaultClass}}"
  ...
  {{on 'focusout' (optional @onFocusOut}}
  {{on 'mouseleave' (optional @onMouseLeave}}
>
  {{yield}}
</Element>
```

This could temporarily be added back in `ember-basic-dropdown` V2 and later removed again in an `ember-basic-dropdown` V3 release when the previously mentioned RFC is landed in Ember.

Happy to help out with the implementation if you agree! 👍

lougreenwood:
Closes #498.


addon/components/basic-dropdown-content.js
```diff
@@ -108,6 +108,17 @@ export default @layout(templateLayout) @tagName('')class BasicDropdownContent ex
...
   // Methods
+  /**
+   * Allows similair behaviour to `ember-composable-helpers`' `optional` helper.
+   * Avoids adding extra dependencies.
+   * Can be removed when the template `V1` compatability event handlers are removed.
+   * 
+   * @see https://github.com/cibernox/ember-basic-dropdown/issues/498
+   * @memberof BasicDropdownContent
+   */
+  @action
```
cibernox: If you are not accessing `this`, there is no need of decorating the function with `@action`

lougreenwood:
@cibernox I just rebased this PR on latest master.

Since I originally created this PR, the repo has moved to TypeScript, which makes pinning this branch in our package.json more difficult as our project is not TS, so we don't have tooling setup to build a TS branch.

With that in mind, I'm keen to pick up this PR and help to get it merged.

So.... I have a question... What do you think about tests for this fix? It seems that proper integration tests would require mocking each of the interaction events to ensure that the callback is properly called. But since this is a stopgap solution, which will probably be un-documented and used by a small number of people and is pretty straight forward, I'm not sure it's worth the effort of adding tests.

WDYT?

lougreenwood:
@cibernox any thoughts on this, would be great to get it merged in so that our (non-TS) project can move to V3 & ember power select 4?

Thanks!

lougreenwood:
@cibernox I've added tests for the temporary event args and we're 🟢.

Would really appreciate a review as this is blocking upgrading our project past 3.16. Thanks!

cibernox:
Merged, but this makes me a bit sad. Clearly the component helper should allow element modifiers to be passed in.

cibernox:
Can you test this using master before I cut a new release?

lougreenwood:
> Merged, but this makes me a bit sad. Clearly the component helper should allow element modifiers to be passed in.
Me too, but I tried to make it easy to remove, so 🤞that something like https://github.com/emberjs/rfcs/issues/497 ever lands.

> Can you test this using master before I cut a new release?
We don't have TypeScript setup for our project, so it will take me some time to handle that. Is there another way to test, maybe cutting some temporary release? We've been using https://github.com/cibernox/ember-basic-dropdown/pull/499/commits/c839061dbf48502c41fbb0d21bd2a986a559e137 for at least 4 months in production now, and all that's changed since then is adding typings and tests...

lougreenwood:
After discussing this on `#e-typescript`, it seems that I shouldn't have any issues using it in a non-TS project, so will test and get back to you 👍

lougreenwood:
I've had a chance to test, it's working well. However, I noticed another related issue which I'll create a PR for on Monday

lougreenwood:
@cibernox Followup PR created - https://github.com/cibernox/ember-basic-dropdown/pull/499 Thanks

lougreenwood:
@cibernox I've been using this branch for about a week now. We haven't done a release on master, but so far in my testing I haven't seen any issues. 👍

PR #501, opened 2019-10-29
Use set when changing `previousVerticalPosition` and `previousHorizontalPosition`.

### LINKED ISSUE #500: Issue with ember 3.13 and `previousHorizontalPosition`
> Error: Assertion Failed: You attempted to update <garaje@component:basic-dropdown::ember1145>.previousHorizontalPosition to "null", but it is being tracked by a tracking context, such as a template, computed property, or observer. In order to make sure the context updates properly, you must invalidate the property when updating it. You can mark the property as `@tracked`, or use `@ember/object#set` to do this.

✖_Security_·_Cool_Chip_SF_·_Envoy

esbanarango:
This PR fixes #500


esbanarango:
@cibernox Let me know what do you think about this. It's currently blocking 3.13 upgrade 🙏 .

PR #502, opened 2019-11-03
Fix development assertion when not rendering in place

raido:
This PR fixes assertion throwing on exception when rendering dropdown menu in place.


cibernox:
Good catch, thanks! I'll release a new version

PR #510, opened 2019-11-21
Fall back to verticalPosition="below" if there isn't enough space either way

### LINKED ISSUE #505: Vertical positioning is off if dropdown is too large for screen
If you have a dropdown with a content that is too large for both above as well as below the trigger, it will always render the dropdown above the trigger. IMHO it should default to below in that case, because the page can extend below, but not above. Right now, if you have a button at the top of the page, it will render a dropdown upwards that will be impossible to read. This is noticeable on mobile devices.

Related to #454

mydea:
If using auto vertical positioning, and there is not enough space either above or below, the dropdown should always be positioned below - this way, the page will at least expand to show the dropdown, whereas it will be cut off if it moves out of the screen above.

Fixes #505

### DIFF
--- a/addon/utils/calculate-position.ts
+++ b/addon/utils/calculate-position.ts
@@ -122,7 +122,9 @@ export let calculateWormholedPosition: CalculatePosition = (trigger, content, de
     let enoughRoomBelow = triggerTopWithScroll + triggerHeight + dropdownHeight < viewportBottom;
     let enoughRoomAbove = triggerTop > dropdownHeight;
 
-    if (previousVerticalPosition === 'below' && !enoughRoomBelow && enoughRoomAbove) {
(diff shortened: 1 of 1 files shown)


mydea:
_ping_ if there is any chance to take a look at this and merge this, that would be great - as this issue makes the dropdown basically unusable on some pages of our app :/

cibernox:
I'd swear this was already the default 🤔 

Anyhow, this makes sense. Approved

cibernox:
Sorry I missed this last week. This was been merged in master, which right now is in beta and requires Ember 3.13 or greater. Can you use the beta or do you need this to be backported to v3?

mydea:
No worries! :) I can use beta, no problem! Thanks for merging :+1:

PR #517, opened 2019-12-20
ignore comment node changes on mutation observer

miguelcobain:
Ember uses comment nodes to keep track of the component boundaries.
I was detecting some unneeded repositions because the mutation observers were detecting these comment nodes insertions/deletions.

I also went ahead and filtered out empty text nodes as well.

`Array.prototype.slice.call` was used to convert a `NodeList` to an array because it is faster than `Array.from()`. Source: https://stackoverflow.com/a/49822981

### DIFF
--- a/addon/components/basic-dropdown-content.js
+++ b/addon/components/basic-dropdown-content.js
@@ -177,7 +177,17 @@ export default @layout(templateLayout) @tagName('')class BasicDropdownContent ex
   @action
   setupMutationObserver(dropdownElement) {
     this.mutationObserver = new MutationObserver((mutations) => {
-      if (mutations[0].addedNodes.length || mutations[0].removedNodes.length) {
+      let shouldReposition = false;
+
+      shouldReposition = Array.prototype.slice.call(mutations[0].addedNodes).some((node) => {
+        return node.nodeName !== '#comment' && !(node.nodeName === '#text' && node.nodeValue === '');
+      });
+
+      shouldReposition = shouldReposition || Array.prototype.slice.call(mutations[0].removedNodes).some((node) => {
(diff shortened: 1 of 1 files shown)


addon/components/basic-dropdown-content.js
```diff
@@ -177,7 +177,15 @@ export default @layout(templateLayout) @tagName('')class BasicDropdownContent ex
   @action
   setupMutationObserver(dropdownElement) {
     this.mutationObserver = new MutationObserver((mutations) => {
-      if (mutations[0].addedNodes.length || mutations[0].removedNodes.length) {
+      let addedNodes = Array.prototype.slice.call(mutations[0].addedNodes).filter((node) => {
```
cibernox: Use `Array#some` instead of `Array#filter` and then length.

miguelcobain: Sounds good. updated. 👍

addon/components/basic-dropdown-content.js
```diff
@@ -177,7 +177,17 @@ export default @layout(templateLayout) @tagName('')class BasicDropdownContent ex
...
     this.mutationObserver = new MutationObserver((mutations) => {
-      if (mutations[0].addedNodes.length || mutations[0].removedNodes.length) {
+      let shouldReposition = false;
+
+      shouldReposition = Array.prototype.slice.call(mutations[0].addedNodes).some((node) => {
+        return node.nodeName !== '#comment' && !(node.nodeName === '#text' && node.nodeValue === '');
+      });
+
+      shouldReposition = Array.prototype.slice.call(mutations[0].removedNodes).some((node) => {
+        return node.nodeName !== '#comment' && !(node.nodeName === '#text' && node.nodeValue === '');
```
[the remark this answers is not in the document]
miguelcobain: Nice catch. Updated.

PR #520, opened 2019-12-23
@shouldReposition function

miguelcobain:
@shouldReposition function can be used to control which mutations should trigger a reposition


cibernox:
@miguelcobain looks great. merged!

PR #529, opened 2020-01-26
Support custom positioning using `bottom` property

seanCodes:
This adds support for handling `style.bottom` if it’s returned from `calculatePosition()`.

`top`, `left` and `right` are already supported, so now with `bottom` we have the complete set! 🙂


cibernox:
Looks good, I'll merge it. Although I'm not sure what the use case is, as using top + height has always been enough for me.

cibernox:
Thanks!

PR #533, opened 2020-02-21
Provide non-default package.json description

reidab:
https://www.emberobserver.com/addons/ember-basic-dropdown uses this in their display and the default blueprint text is confusing.


cibernox:
Thanks!

PR #541, opened 2020-03-31
Fix animationEnabled getter to account for false value

mehulkar:
If `@animationEnabled={{false}}` is passed, the getter will not use it

### DIFF
--- a/addon/components/basic-dropdown-content.js
+++ b/addon/components/basic-dropdown-content.js
@@ -62,7 +62,7 @@ export default @layout(templateLayout) @tagName('')class BasicDropdownContent ex
   // CPs
   @computed
   get animationEnabled() {
-    if ('_animationEnabled' in this) {
+    if (this._animationEnabled !== 'undefined') {
       return this._animationEnabled;
     }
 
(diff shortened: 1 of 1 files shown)


mehulkar:
cc @miguelcobain since you did some recent work in this area also

miguelcobain:
@mehulkar I think a better check would be:

```js
if (this._animationEnabled !== undefined) {
  return this._animationEnabled;
}
```

because, if for some reason we happen to pass `@animationEnabled={{this.someUndefinedProperty}}` to the component, the value will be `undefined`, and I think in that case we still want the default behaviour and not to bail out of animations.

cibernox:
I agree with @miguelcobain, checking undefined is probably better because just blindly forwarding attributes from a wrapper component is a common patterns and in those cases you probably want the default behavior to kick in.

mehulkar:
@miguelcobain @cibernox good idea, updated

mehulkar:
Thanks. I see a 2.0.14 tag on GH now, but not on npm. Is that going to get published soon?

cibernox:
It turns out I lost my credentials and I can't recover them because i have
2fa and I have recently changed phones, so can't publish to npm right now.
I am in contact with support to get access again. Can you use the github
tag for now? Hopefully won't take more than a day.

miguelcobain:
Wait, am I seeing correctly?

Seems like we're doing
```js
this._animationEnabled !== 'undefined'
```

instead of 

```js
this._animationEnabled !== undefined
```

?

mehulkar:
🤦 @miguelcobain you're absolutely right. Should have been `typeof`. I upgraded to the glimmer version so I never saw this break

PR #554, opened 2020-05-22
use default values if a property (that can come from an argument) has a value that is `undefined`

emattias:
This makes it so that the default values are used even if arguments are sent in but if their values are `undefined` the defaults will be used. 

For example `ember-power-select` [always sends some arguments](https://github.com/cibernox/ember-power-select/blob/master/addon/templates/components/power-select.hbs#L2) to `BasicDropdown` which caused the default values to not be used.

This should fix that.

(This pr is done for [v2 branch](https://github.com/cibernox/ember-basic-dropdown/tree/v2) since we cant upgrade to v3 yet because we use https://github.com/cibernox/ember-power-select-with-create which requires `ember-basic-dropdown` v2)

### DIFF
--- a/addon/components/basic-dropdown.js
+++ b/addon/components/basic-dropdown.js
@@ -20,6 +20,13 @@ const ignoredStyleAttrs = [
   'height'
 ];
 
+const defaultValues = {
+  renderInPlace: false,
+  verticalPosition: 'auto', // above | below
+  horizontalPosition: 'auto', // auto-right | right | center | left
+  matchTriggerWidth: false,
+}
+
 export default @layout(templateLayout) @tagName('') class BasicDropdown extends Component {
   top = null;
   bottom = null;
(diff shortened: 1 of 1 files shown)


cibernox:
Just to check, did you try v3? I think that in V3 it's solved.

Not that I can't merge it in v2, tho.

emattias:
> Just to check, did you try v3? I think that in V3 it's solved.
> 
> Not that I can't merge it in v2, tho.

I added this while you wrote your comment :) -> 
> This pr is done for v2 branch  since we cant upgrade to v3 yet because we use https://github.com/cibernox/ember-power-select-with-create which requires ember-basic-dropdown v2

cibernox:
@emattias thanks for pointing that up. I've update ember-power-select-with-create to EPS 4.0 and ember-basic-dropdown 3.0, so you should be able to update now.

If you can, I'd rather leave v2 alone.

emattias:
> @emattias thanks for pointing that up. I've update ember-power-select-with-create to EPS 4.0 and ember-basic-dropdown 3.0, so you should be able to update now.
> 
> If you can, I'd rather leave v2 alone.

Thanks but we cant update to 3.0 right now unfortunately. Either we use the fork or if you can, please merge these changes.

PR #556, opened 2020-05-28
Issue 555 - Always set aria-expanded, also when collapsed.

### LINKED ISSUE #555: aria-expanded must be set to "false" if not expanded
I found this in the [Ghost](https://ghost.org/) admin panel, for example the Publish button, or the More item on the mobile navigation bar. If the content is collapsed, aria-expanded is not set at all on the trigger. But this means that it is not communicating to screen readers that this is expandable. So, if the content is collapsed, aria-expanded must still be set, but to "false". The relevant HBS line is here:

https://github.com/cibernox/ember-basic-dropdown/blob/master/addon/templates/components/basic-dropdown-trigger.hbs#L8

Question: Is it enough to just set a "false" string as a second string before the closing double-braces? And then adjust the tests obviously. :-)

MarcoZehe:
aria-expanded is a property that must always be set when used, to either `"true"` when expanded, or `"false"` if collapsed. Omitting it when collapsed means that assistive technologies aren't aware that this element is expandable.

Closes #555 .

### DIFF
--- a/addon/templates/components/basic-dropdown-trigger.hbs
+++ b/addon/templates/components/basic-dropdown-trigger.hbs
@@ -5,7 +5,7 @@
     tabindex={{unless @dropdown.disabled "0"}}
     data-ebd-id="{{@dropdown.uniqueId}}-trigger"
     aria-owns="ember-basic-dropdown-content-{{@dropdown.uniqueId}}"
-    aria-expanded={{if @dropdown.isOpen "true"}}
+    aria-expanded={{if @dropdown.isOpen "true" "false"}}
     aria-disabled={{if @dropdown.disabled "true"}}
     {{will-destroy this.removeGlobalHandlers}}
     ...attributes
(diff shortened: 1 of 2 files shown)


addon/templates/components/basic-dropdown-trigger.hbs
```diff
@@ -5,7 +5,7 @@
     tabindex={{unless @dropdown.disabled "0"}}
     data-ebd-id="{{@dropdown.uniqueId}}-trigger"
     aria-owns="ember-basic-dropdown-content-{{@dropdown.uniqueId}}"
-    aria-expanded={{if @dropdown.isOpen "true"}}
+    aria-expanded={{if @dropdown.isOpen "true" "false"}}
```
cibernox, line 8: There's a slightly less verbose way of doing this: 
```suggestion
    aria-expanded="{{@dropdown.isOpen}}
```

MarcoZehe:
Thank you! I am still learning hbs, so wasn't aware that these directly translate to the string literals. Your change looks even better than mine. ;-)

PR #558, opened 2020-05-31
Use aria-owns on the trigger's container instead of on the trigger

### LINKED ISSUE #557: aria-owns should not be used on the trigger, but on its parent
This was originally reported as https://github.com/TryGhost/Ghost/issues/11868.

The problem is that, right now, aria-owns gets set on the same element that the button role is on. This makes the basic-dropdown-content container a child of the button. But buttons by definition can only have text content, or graphics, but no other buttons, links, or other inputs inside it.

Some screen readers like NVDA are quite forgiving, although the announcement of everything being a button in addition to a text field or such, is still annoying. Other screen readers, like Orca on Linux, are not so forgiving, as can be seen in the report against the Ghost project.

I am using a local branch where I try to replace aria-owns with aria-controls to still have a relation that screen readers may use, but in the dummy application at least, content is not being added below that, even though the markup says so, but appended to the end of the document to a basic-dropdown-wormhole div that seems to always be there.

So to always make the dropdown content a sibling of the trigger, aria-owns should be applied to the *parent* of the trigger. If only the ID of the content is specified, what will happen is that first all regular children get added to the accessibility tree as usual, and then the aria-owned ones appended to the end. And this is what we want.

Question is: How should I best do that? @cibernox any suggestions?

MarcoZehe:
Closes #557.

aria-owns should not be used to make the content elements children of the button, since this is invalid accessibility structure. Screen readers like Orca on Linux don't handle this well.

Instead, use aria-controls to build a relationship between the trigger and content, and add aria-owns to the parent of the trigger so the content gets appended as the last accessible child of that container.

### DIFF
--- a/addon/components/basic-dropdown-content.ts
+++ b/addon/components/basic-dropdown-content.ts
@@ -343,7 +343,7 @@ function dropdownIsValidParent(el: Element, dropdownId: string): boolean {
     return false;
   } else {
     let closestAttrs = closestDropdown.attributes as unknown as any
-    let trigger = document.querySelector(`[aria-owns=${closestAttrs.id.value}]`);
+    let trigger = document.querySelector(`[aria-controls=${closestAttrs.id.value}]`);
     if (trigger === null) return false;
     let parentDropdown = closestContent(trigger);
     if (parentDropdown === null) return false;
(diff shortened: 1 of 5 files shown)


mkszepp:
Not sure when it was done, but we have both aria properties on trigger already and it looks like its a11y conform

PR #569, opened 2020-07-21
#498 - Allow overriding built-in event handlers

lougreenwood:
Followup to https://github.com/cibernox/ember-basic-dropdown/pull/499

During testing https://github.com/cibernox/ember-basic-dropdown/pull/499, I noticed that it's not possible to override the behaviour of the built in event handlers. For example, to make the dropdown only appear on hover instead of click.

This PR allows passing any of the following args to override the built in event handlers instead of augment them (as https://github.com/cibernox/ember-basic-dropdown/pull/499 previously allowed):

- `@onClick`
- `@onKeyDown`
- `@onMouseDown`
- `@onTouchEnd`
- `@onTouchStart`

It's still possible to augment the existing event handlers by using the `{{on ...}}` modifier (however, this is not available when yielding an instance of the EBD Trigger using the component helper).

### DIFF
--- a/addon/templates/components/basic-dropdown-trigger.hbs
+++ b/addon/templates/components/basic-dropdown-trigger.hbs
@@ -8,24 +8,24 @@
     aria-expanded="{{@dropdown.isOpen}}"
     aria-disabled={{if @dropdown.disabled "true"}}
     {{will-destroy this.removeGlobalHandlers}}
-    ...attributes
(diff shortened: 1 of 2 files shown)


lougreenwood:
I managed to work around this another way, so this PR can be closed, but it does highlight a problem that the `{{on ...}}` modifiers can't be used to override other `{{on}}` modifiers :(

Issue #574, opened 2020-08-13
Object.assign -> assign from '@ember/polyfills' for IE11


raido:
This line breaks for IE11, due to using `Object.assign` instead of `assign` as polyfill from Ember itself.

https://github.com/cibernox/ember-basic-dropdown/blob/bcb9a3919e3d4aa59583e9fabc57f676afe1d189/addon/components/basic-dropdown.ts#L221

raido:
I'll open a PR for this.

PR #575, opened 2020-08-13
Use assign polyfill from Ember instead Object.assign for IE11

raido:
Closes https://github.com/cibernox/ember-basic-dropdown/issues/574


cibernox:
Thanks!

PR #579, opened 2020-09-09
Fix horizontalPosition="center" when using verticalPosition="above" & renderInPlace="true"

danielfinke:
`horizontalPosition="center"` was being ignored if you were using both `verticalPosition="above"` and `renderInPlace="true"`.

`positionData.style` was being overwritten in the `verticalPosition === 'above'` block - only the `top` prop should be modified


cibernox (approved):

cibernox:
Thanks!

PR #587, opened 2020-11-19
[Embroider] replace with safe component

akatz:
This replaces unsafe component usage with the embroider util that allows  EBD to reach `Optimized Embroider Safe` support level as per https://github.com/embroider-build/embroider/blob/master/ADDON-AUTHOR-GUIDE.md


ef4:
Thanks for working on this, I was about to do it and found your PR.

I'm curious about: https://github.com/cibernox/ember-basic-dropdown/pull/587/commits/88a2555db4ba885e6fb6acfd1a0eccc882cce081

Were you seeing a problem without this change? Because as far as I know, embroider wouldn't require it. 

I'm going to build off this PR and add test coverage.

akatz:
@ef4 So if I remember correctly ( and I might just not actually because sadly I'm trying a lot of things without actually knowing exactly how embroider actually works ) it seemed like it worked without it. But I had issues actually using it in my app ( I installed this branch locally with the ember-element-helper embroider branch ). I believe starting the ember server ( as well as `yarn embroider-compat-audit` both warned me that the other syntax was not statically analyze-able.

akatz:
@ef4 so if I misunderstand something and the other syntax works, that really help as I introduced a duplication in a few key places to try and work around the issue.

alexborovkov:
@cibernox Hello! Any thoughts when this PR will be merged? I've got compatibility error with embroider and this should solve it.

cibernox:
Sorry for neglecting this for so long. Checking it now.

addon/templates/components/basic-dropdown-content.hbs
```diff
@@ -28,6 +29,31 @@
           {{yield}}
         </Element>
       {{/let}}
+      {{else}}
```
cibernox, line 32: It's a shame that we have to repeat the block. @ef4 is there a less verbose way of making the embroider gods happy?

cibernox (approved):

cibernox:
It's a shame we have that repetition, but it's better than not working in embroider.

PR #591, opened 2020-12-18
CSP-safe positioning

meirish:
Currently the positioning of the basic dropdown content requires you to use an `unsafe-inline` directive if you're using a [CSP](https://developer.mozilla.org/en-US/docs/Web/HTTP/CSP). This PR changes the positioning of the content from an inline, html-safe string to using https://github.com/jelhan/ember-style-modifier which adds the styles in a CSP-friendly way.

I'm not sure if you're interested in this change as it brings in a few deps, and I'm not sure what your compatibility story is. I think it's worth it and I believe the change is non-breaking.

Thanks!

### DIFF
--- a/addon/components/basic-dropdown-content.ts
+++ b/addon/components/basic-dropdown-content.ts
@@ -3,7 +3,6 @@ import Component from '@glimmer/component';
 import { tracked } from '@glimmer/tracking';
 import { join } from '@ember/runloop';
 import { getOwner } from '@ember/application';
-import { htmlSafe } from '@ember/string';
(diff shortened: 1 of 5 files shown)


cibernox:
Sorry, I missed this PR completely.

TBH, i kind of like it better. It's cleaner that the intent is clear in the template.

Merged!

cibernox:
Sorry, I missed this PR completely.

TBH, i kind of like it better. It's cleaner that the intent is clear in the template.

Merged!

meirish:
awesome, thanks @cibernox !

PR #594, opened 2021-02-16
Install ember-cli-memory-leak-detector

steveszc:
~Congrats! EBD has no memory leaks (at least no memory leaks caused by codepaths exercised by the test suite).~
Oops I just noticed that CI does not appear to be running, but the test suite did not fail locally with this new addon installed

This PR installs an add-on I authored called ember-cli-memory-leak-detector.

The add-on works by recording a heap snapshot after the suite tests finishes running. The heap snapshot is then searched for any retained instances of the host app/addon's ES classes (indicating a memory leak). If it finds retained classes in the heap snapshot, the test suite fails and a report of the retained classes is logged in the console output.

After installing ember-cli-memory-leak-detector we can confirm that EBD has no memory leaks, and we can ensure that no memory leak is ever introduced into EBD in the future, which is a big win for a such a popular addon!

I'll be giving a talk at EmberConf this year about ember-cli-memory-leak-detector and I'd like to get it installed in at least one or two popular addons prior to giving the talk. EBD is a great candidate because it is popular, already using ES classes, and has no memory leaks. If ember-cli-memory-leak-detector gets traction from ember conf we can move the ember community toward eliminating memory leaks altogether!

### DIFF
--- a/package.json
+++ b/package.json
@@ -39,6 +39,7 @@
     "ember-cli-dependency-checker": "^3.2.0",
     "ember-cli-fastboot": "^2.0.0",
     "ember-cli-inject-live-reload": "^2.0.2",
+    "ember-cli-memory-leak-detector": "^0.5.1",
     "ember-cli-sass": "^10.0.1",
     "ember-cli-typescript-blueprints": "^3.0.0",
     "ember-cli-uglify": "^3.0.0",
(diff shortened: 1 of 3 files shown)


cibernox:
That's a very nice idea. In fact this addon DID have memory leaks in the past and those where tricky to find (but easy to fix).

Merging it. And I have to migrate the test suite to github actions to fix CI.

PR #595, opened 2021-02-16
use style modifier in the content else block too

meirish:
In  #591 I aimed to make this work with a strict CSP, and it did for the most part. We got around to using it (via ember-power-select) and were still seeing a few errors.

This adds the same `{{style}}` modifier usage to the `<Element>` rendered if `@htmlTag` is passed.
The initial opening now uses `element.style.setProperty` instead of `element.setAttribute` which is not CSP-safe.


cibernox:
Sorry for the delay, I'm pretty busy lately and I try to binge-review all my github issues and PRs on Fridays. 

Thanks!

PR #602, opened 2021-03-04
Use `@embroider/macros` to conditionally import `@ember/test-helpers`

mydea:
The current way to conditionally import `@ember/test-helpers/dom/get-root-element` breaks in an app using embroider with:

```
staticAddonTestSupportTrees: true
```

This PR refactors this to instead use `@embroider/macros`, which can be statically analyzed. Functionally, nothing should change, I think. I tried it in an app and it worked fine.

### DIFF
--- a/addon/components/basic-dropdown-content.ts
+++ b/addon/components/basic-dropdown-content.ts
@@ -2,7 +2,6 @@ import { action } from "@ember/object";
 import Component from '@glimmer/component';
 import { tracked } from '@glimmer/tracking';
 import { join } from '@ember/runloop';
-import { getOwner } from '@ember/application';
 import { getScrollParent } from '../utils/calculate-position';
 import {
   distributeScroll,
@@ -11,6 +10,7 @@ import {
 } from '../utils/scroll-helpers';
 import hasMoved from '../utils/has-moved';
 import { Dropdown } from './basic-dropdown';
+import { isTesting } from '@embroider/macros';
 
 interface Args {
   transitioningInClass?: string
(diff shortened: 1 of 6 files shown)


addon/components/basic-dropdown-content.ts
```diff
@@ -54,8 +54,7 @@ export default class BasicDropdownContent extends Component<Args> {
   }
 
   get animationEnabled(): boolean {
-    let config = getOwner(this).resolveRegistration('config:environment');
-    return config.environment !== 'test';
+    return !isTesting();
```
cibernox, line 57: What is this witchcraft! I like it

addon/components/basic-dropdown.ts
```diff
@@ -279,28 +277,32 @@ export default class BasicDropdown extends Component<Args> {
...
-            return requirejs('@ember/test-helpers/dom/get-root-element').default().id as string;
-          } catch(ex) {
-            // no op
-          }
-        }
-        let rootView = document.querySelector('#ember-testing > .ember-view');
+    let config = getOwner(this).resolveRegistration("config:environment");
+
+    // This takes care of stripping this code out if not running tests
+    if (macroCondition(isTesting())) {
```
cibernox, line 283: Does this work both when running `ember test` and also when running the dev server but visiting the `/tests` route?

cibernox (approved):

addon/components/basic-dropdown.ts
mydea: Yeah, that is actually quite cool, it will even work when e.g. running `ember test --environment=production`.

cibernox:
merged! I'll release a new version today

cibernox:
Published in 3.0.15 🎉

PR #604, opened 2021-03-08
Extract optional tag into component to avoid repetition

mydea:
By extracting the code to build an HTML element with an optional htmlTag into a component, we can avoid duplicating the whole template code for both the trigger and the content.


addon/components/basic-dropdown-optional-tag.ts
```diff
@@ -0,0 +1,5 @@
+import Component from '@glimmer/component';
+
+interface BasicDropdownOptionalTagArgs {}
+
+export default class BasicDropdownOptionalTag extends Component<BasicDropdownOptionalTagArgs> {}
```
cibernox, line 5: Could this be a template-only component?

mydea: I wasn't quite sure how that would work in an app that hasn't enabled octane? would it then not be a "regular" ember-component with a wrapping element?

cibernox: I didn't consider apps with octane disabled. This looks good.

cibernox (approved):

cibernox:
Thanks!

Issue #607, opened 2021-03-24
ember-basic-dropdown 3.0.16 doesn't work with some versions of Ember <= 3.24


pgengler:
Version 3.0.16 of the addon doesn't seem to work with some versions of Ember that are > 3.13 (which is part of the ember-try config) and < 3.25 (which aren't). You can see this reflected in the version compatibility report on Ember Observer (the output of the tests that ran to produce that result are here: https://gist.github.com/pgengler/03b464a76ea7e60d19d4f3fb3d996e07), and I've verified it both by including it in an app using Ember 3.24 and by pulling down the source and running ember-try with Ember 3.24.

The scenarios currently in the ember-try config (Ember 3.13, release [3.25], beta [3.26], and canary [3.27]) all pass, but the versions in between seem to fail (3.16, 3.20, 3.24).

This is one example of the error:
```
not ok 1 Chrome 89.0 - [97 ms] - Integration | Component | basic-dropdown: Its `toggle` action opens and closes the dropdown
    ---
        actual: >
            [object Object]
        stack: >
            TypeError: Cannot read property 'length' of undefined
                at meta (http://localhost:7357/assets/vendor.js:45694:34)
                at compilable (http://localhost:7357/assets/vendor.js:45199:57)
                at TemplateImpl.asLayout (http://localhost:7357/assets/vendor.js:46535:28)
                at Object.evaluate (http://localhost:7357/assets/vendor.js:51081:52)
                at AppendOpcodes.evaluate (http://localhost:7357/assets/vendor.js:49062:19)
                at LowLevelVM.evaluateSyscall (http://localhost:7357/assets/vendor.js:52563:22)
                at LowLevelVM.evaluateInner (http://localhost:7357/assets/vendor.js:52519:14)
                at LowLevelVM.evaluateOuter (http://localhost:7357/assets/vendor.js:52511:14)
                at VM.next (http://localhost:7357/assets/vendor.js:53636:24)
                at VM._execute (http://localhost:7357/assets/vendor.js:53620:23)
        message: >
            Cannot read property 'length' of undefined
        negative: >
            false
        browser log: |
            ERROR: 
            
            Error occurred:
            
            - While rendering:
              -top-level
                application
                  index
```

pgengler:
I guess this was a flake (or a transitive dependency that broke/got fixed in between) - ran this again both locally and on Ember Observer and this time it passed with these versions.

PR #609, opened 2021-03-27
Add current Ember LTS versions (3.20 and 3.24) to ember-try config

pgengler:
This updates the ember-try config to include the current (as of this PR) LTS versions of Ember (in addition to the existing scenarios of Ember 3.13, current release, beta, and canary). I noticed in #607 that the tests were failing with versions of Ember between 3.16 and 3.24, while they passed with 3.13 and 3.25.

This will require some effort going forward to keep the LTS scenarios in sync with the current LTS versions of Ember but they don't change very often so there won't be a lot of churn on that account.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -72,6 +72,8 @@ jobs:
       matrix:
         ember-try-scenario:
           - ember-3.13
+          - ember-lts-3.20
+          - ember-lts-3.24
           - ember-release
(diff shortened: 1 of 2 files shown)


pgengler:
The original motivation behind this is gone (tests are passing & the addon is working with the relevant versions) so I'm going to close this - it might still have some value so feel free to reopen/merge if you want.

PR #611, opened 2021-04-06
Allow ember-element-helper ^0.4.0

mydea:
To allow host apps to update this dependency. There aren't really any breaking changes as far as I can see (see dependabot screenshot from one of our apps below):

Screenshot from 2021-04-06 09-26-57


lukemelia:
If we're going to take this approach, it would be great to allow ember-element-helper 0.5.0, which eliminates a warning when used with embroider.

mydea:
This has been updated to 0.5.0 by now anyhow, so closing this!

Issue #617, opened 2021-05-31
Does not work inside Shadow DOM


lolmaus:
Hi!

We use Shadow DOM to allow embedding our app into any website. It solves the following problems of `<iframe>`:

* Modal backdrop and sidebar modals are rendered to the whole viewport and not contained in a smaller rectangle.
* Host website styles do not leak into the app, do not deface it and do not require ugly overrides.

Shadow DOM, in the absence of alternative solutions, is crucial to our business.

Ember works flawlessly in Shadow DOM, it does not require any modifications (beyond embedding itself, for which we have a solution: https://github.com/kaliber5/ember-embedded-snippet/ ). But Shadow DOM breaks all addons that rely on things like `document.getElementById` which fails to look up elements inside Shadow DOM.

`ember-basic-dropdown` and, by extension, `ember-power-select` are the biggest offenders.

CC @simonihmig

BlueCutOfficial:
Hi!
I join the `#needShadowDOM` team ^^ 
We also want to use Shadow DOM to serve our app as a Web Component, so we face the same issue as we rely on `ember-basic-dropdown` for all our datepickers, calendar or other selects.

mkszepp:
closed with #791

PR #618, opened 2021-05-31
Shadow DOM support, implements #617

lolmaus:
This is a proof-of-concept PR that implements basic Shadow DOM support, closing #617. Please see the issue for explanation of why Shadow DOM may be necessary.

This PR misses Shadow DOM-specific tests and documentation. I would like to receive a general approval and guidelines from `ember-basic-dropdown` maintainers before investing moore time into it.

Existing tests pass, and we already use this branch in a project.

This PR does not cover all possible cases of arbitrarily creating Shadow DOM containers and arbitrarily positioning the wormhole container. Supporting all such cases would be tricky and would like require a generic hack like https://github.com/Georgegriff/query-selector-shadow-dom which is undesirable.

But this PR does cover the most basic and demaded case: wrapping your whole Ember app with one Shadow DOM container. It can be done effortlessly with our https://github.com/kaliber5/ember-embedded-snippet/  addon.

In order to avoid leaking app styles into the host website, stylesheets are referenced from inside Shadow DOM. This means that a wormholed dropdown content would appear unstyled if the wormhole is outside of the Shadow DOM.

Thus, a requirement of using `ember-basic-dropdown` in Shadow DOM is to set up a wormhole container in `application.hbs` and override `['ember-basic-dropdown'].destination` to prevent the addon from emitting a default one into `{{content-for "body-footer"}}`.

CC @simonihmig

### DIFF
--- a/addon/components/basic-dropdown-content.ts
+++ b/addon/components/basic-dropdown-content.ts
@@ -11,6 +11,7 @@ import {
 import hasMoved from '../utils/has-moved';
 import { Dropdown } from './basic-dropdown';
 import { isTesting } from '@embroider/macros';
+import { ref } from 'ember-ref-bucket';
 
 interface Args {
   transitioningInClass?: string
@@ -28,6 +29,7 @@ interface Args {
   width: string | undefined
   height: string | undefined
   otherStyles: Record<string, string>
+  triggerElement: HTMLElement | undefined
   onFocusIn?: (dropdown?: Dropdown, event?: FocusEvent) => void
   onFocusOut?: (dropdown?: Dropdown, event?: FocusEvent) => void
   onMouseEnter?: (dropdown?: Dropdown, event?: MouseEvent) => void
@@ -54,7 +56,7 @@ export default class BasicDropdownContent extends Component<Args> {
   }
 
   get animationEnabled(): boolean {
-    return !isTesting(); 
+    return !isTesting();
   }
 
   /**
@@ -69,11 +71,10 @@ export default class BasicDropdownContent extends Component<Args> {
 
   @action
   setup(dropdownElement: Element): void {
-    let triggerElement = document.querySelector(`[data-ebd-id=${this.args.dropdown.uniqueId}-trigger]`);
     this.handleRootMouseDown = (e: MouseEvent | TouchEvent): any => {
-      if (e.target === null) return;
-      let target = e.target as Element;
-      if (hasMoved(e as TouchEvent, this.touchMoveEvent) || dropdownElement.contains(target) || triggerElement && triggerElement.contains(target)) {
+      let target = (e?.composedPath()[0] || e.target) as Element;
+      if (target === null) return;
+      if (hasMoved(e as TouchEvent, this.touchMoveEvent) || dropdownElement.contains(target) || this.args.triggerElement && this.args.triggerElement.contains(target)) {
         this.touchMoveEvent = undefined;
         return;
       }
@@ -93,8 +94,8 @@ export default class BasicDropdownContent extends Component<Args> {
       document.addEventListener('touchstart', this.touchStartHandler, true);
       document.addEventListener('touchend', this.handleRootMouseDown, true);
     }
-    if (triggerElement !== null) {
-      this.scrollableAncestors = getScrollableAncestors(triggerElement);
+    if (this.args.triggerElement !== null) {
+      this.scrollableAncestors = getScrollableAncestors(this.args.triggerElement);
     }
     this.addScrollHandling(dropdownElement);
   }
@@ -194,9 +195,10 @@ export default class BasicDropdownContent extends Component<Args> {
   addScrollHandling(dropdownElement: Element): void {
     if (this.args.preventScroll === true) {
       let wheelHandler = (event: WheelEvent) => {
-        if (event.target === null) return;
-        let target = event.target as Element;
-        if (dropdownElement.contains(target) || dropdownElement === event.target) {
+        let target = (event.composedPath?.()[0] || event.target) as Element;
+
+        if (target === null) return;
+        if (dropdownElement.contains(target) || dropdownElement === target) {
           // Discover the amount of scrollable canvas that is within the dropdown.
           const availableScroll = getAvailableScroll(target, dropdownElement);
 
--- a/addon/components/basic-dropdown-content.ts
+++ b/addon/components/basic-dropdown-content.ts
@@ -72,7 +72,7 @@ export default class BasicDropdownContent extends Component<Args> {
   @action
   setup(dropdownElement: Element): void {
     this.handleRootMouseDown = (e: MouseEvent | TouchEvent): any => {
-      let target = (e?.composedPath()[0] || e.target) as Element;
+      let target = (e.composedPath?()[0] || e.target) as Element;
       if (target === null) return;
       if (hasMoved(e as TouchEvent, this.touchMoveEvent) || dropdownElement.contains(target) || this.args.triggerElement && this.args.triggerElement.contains(target)) {
         this.touchMoveEvent = undefined;
--- a/addon/components/basic-dropdown-content.ts
+++ b/addon/components/basic-dropdown-content.ts
@@ -72,7 +72,7 @@ export default class BasicDropdownContent extends Component<Args> {
   @action
   setup(dropdownElement: Element): void {
     this.handleRootMouseDown = (e: MouseEvent | TouchEvent): any => {
-      let target = (e.composedPath?()[0] || e.target) as Element;
+      let target = (e.composedPath?.()[0] || e.target) as Element;
       if (target === null) return;
       if (hasMoved(e as TouchEvent, this.touchMoveEvent) || dropdownElement.contains(target) || this.args.triggerElement && this.args.triggerElement.contains(target)) {
         this.touchMoveEvent = undefined;
--- a/addon/components/basic-dropdown-content.ts
+++ b/addon/components/basic-dropdown-content.ts
@@ -11,7 +11,7 @@ import {
 import hasMoved from '../utils/has-moved';
 import { Dropdown } from './basic-dropdown';
 import { isTesting } from '@embroider/macros';
-import { ref } from 'ember-ref-bucket';
+import { getOwner } from '@ember/application';
 
 interface Args {
   transitioningInClass?: string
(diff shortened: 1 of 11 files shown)


addon/components/basic-dropdown-content.ts
```diff
@@ -50,11 +52,11 @@ export default class BasicDropdownContent extends Component<Args> {
 
 
   get destinationElement(): Element | null {
-    return document.getElementById(this.args.destination);
+    return document.getElementById(this.args.destination) ?? getOwner(this).rootElement.querySelector?.(`[id="${this.args.destination}"]`);
```
lolmaus, line 55: There's a little bit of code duplication with the `getOwner(this).rootElement` part. Not sure if it's worth extracting into util.

addon/components/basic-dropdown-content.ts
```diff
@@ -69,11 +71,10 @@ export default class BasicDropdownContent extends Component<Args> {
 
   @action
   setup(dropdownElement: Element): void {
-    let triggerElement = document.querySelector(`[data-ebd-id=${this.args.dropdown.uniqueId}-trigger]`);
     this.handleRootMouseDown = (e: MouseEvent | TouchEvent): any => {
-      if (e.target === null) return;
-      let target = e.target as Element;
-      if (hasMoved(e as TouchEvent, this.touchMoveEvent) || dropdownElement.contains(target) || triggerElement && triggerElement.contains(target)) {
+      let target = (e?.composedPath()[0] || e.target) as Element;
```
lolmaus, line 75: `event.target` does not cross the Shadow DOM border: if the clicked element is inside Shadow DOM, then `event.target` will point at the Shadow DOM root, and the `this.args.triggerElement.contains(target)` check will fail incorrectly.

`event.composedPath()[0]` is equivalent to `event.target` but does reference the clicked element inside Shadow DOM root.

addon/components/basic-dropdown-trigger.ts
```diff
@@ -107,24 +109,29 @@ export default class BasicDropdownTrigger extends Component<Args> {
...
-        e.target.dispatchEvent(event);
+        target.dispatchEvent(event);
       }
     }, 0);
     e.preventDefault();
   }
 
+  @action
+  registerTriggerElement(triggerElement: HTMLElement): void {
+    this.args.onInsert?.(triggerElement);
```
lolmaus, line 132: This passes a reference of trigger element back into the parent `basic-dropdown` component.

Essentially, it is a lightweight equivalent of https://github.com/lifeart/ember-ref-modifier . `ember-ref-modifier` is deprecated in favor of https://github.com/lifeart/ember-ref-bucket , but `ember-ref-bucket` does not work well for our case (the reference needs to be created in the child component and passed to parent).

addon/utils/calculate-position.ts
```diff
@@ -35,6 +35,12 @@ export let calculateWormholedPosition: CalculatePosition = (trigger, content, de
   let anchorPosition = window.getComputedStyle(anchorElement).position;
   while (anchorPosition !== 'relative' && anchorPosition !== 'absolute' && anchorElement.tagName.toUpperCase() !== 'BODY') {
     anchorElement = anchorElement.parentNode as HTMLElement;
+
+    if ((anchorElement instanceof DocumentFragment) && anchorElement.host) {
+      // Passing through the Shadow DOM border
+      anchorElement = anchorElement.host;
```
lolmaus, line 41: 👀

tests/unit/utils/scroll-helpers-test.js
```diff
@@ -70,12 +70,12 @@ test('distributeScroll', function (assert) {
...
-  assert.strictEqual(container.scrollTop, 10);
-  assert.strictEqual(child.scrollTop, 0);
+  assert.strictEqual(Math.round(container.scrollTop), 10);
+  assert.strictEqual(Math.round(child.scrollTop), 0);
 
   distributeScroll(0, 40, grandchild, container);
-  assert.strictEqual(container.scrollTop, 10);
-  assert.strictEqual(child.scrollTop, 40);
+  assert.strictEqual(Math.round(container.scrollTop), 10);
+  assert.strictEqual(Math.round(child.scrollTop), 40);
```
lolmaus, line 78: This test was failing for me locally with an error like `expected 10 to be equal to 9.6897234502346`. I believe, this test failure is unrelated to the PR.

apellerano-pw:
Thank you for this work. My team has been maintaining an internal fork of the 1.x line with similar shadow DOM support. (We only recently became Octane-ready.)

We were able to upgrade to this fork with no issues.

addon/components/basic-dropdown-content.ts
apellerano-pw, line 55: The use of nullish coalescing operator `??` here means there is a new dependency on either `@babel/preset-env` or `@babel/plugin-proposal-nullish-coalescing-operator`.

If the consuming app doesn't already have that plugin, it will fail during build:

```
Build Error (UglifyWriter)

Unexpected token: operator (?)
```

apellerano-pw: Actually I don't know what the root of the problem was... this turned into a little adventure for me. Had to upgrade from ember-cli-uglify 2.x to ember-cli-terser 4.x before it finally would build.

I'm sure it's something specific to my project's dependencies and not something this addon has to worry about. However, since this would be the first usage of `??` in this addon, maybe it's best to dodge the problem entirely and avoid using `??`.

simonihmig: Glad you figured it our already!

This issue comes up quite a lot, but always boils down to either not having `config/target.js` not configured correctly, or using uglify, or both. `ember-cli-babel` already brings all what is need to correctly transpile it. 

UglifyJS is not maintained anymore AFAIK, and does not support latest ES features. So chokes on nullish coalescing and other recent language features when you haven't IE or other older browsers in your `target.js` causing Babel to not transpile those away. 

So switching to `ember-cli-terser` was the right thing to do, which is part of the app/addon boilerplate already for a long time. 

tl;dr: using latest ES features here should be perfectly fine IMHO!

apellerano-pw: Ah, thank you for the explanation! I feel much better about going down that rabbit hole now.

apellerano-pw:
This PR introduces `ember-ref-bucket` to the package.json but doesn't look to be using it. The `npm install` also may have been run with npm 7 even though this lib is still using npm 6, as it upgrades the entire package-lock.json to the v2 lockfile format.

Could these npm changes be divorced from the rest of the shadow dom implementation? That might make it easier for upstream to accept the PR.

addon/components/basic-dropdown-content.ts
espern: I think that the event is always defined, but `composedPath` isn't
```suggestion
      let target = (e.composedPath?()[0] || e.target) as Element;
```

lolmaus: A-a-and this broke the build and the apps that depended on this branch. 😒

addon/components/basic-dropdown-content.ts
```diff
@@ -69,11 +71,10 @@ export default class BasicDropdownContent extends Component<Args> {
 
   @action
   setup(dropdownElement: Element): void {
-    let triggerElement = document.querySelector(`[data-ebd-id=${this.args.dropdown.uniqueId}-trigger]`);
     this.handleRootMouseDown = (e: MouseEvent | TouchEvent): any => {
-      if (e.target === null) return;
-      let target = e.target as Element;
-      if (hasMoved(e as TouchEvent, this.touchMoveEvent) || dropdownElement.contains(target) || triggerElement && triggerElement.contains(target)) {
+      let target = (e.composedPath?()[0] || e.target) as Element;
```
lolmaus, line 75:
```suggestion
      let target = (e.composedPath?.()[0] || e.target) as Element;
```

CC @espern

espern: oops, my bad.
of course 👍🏻

herzzanu:
@lolmaus any luck we can push this forward? Do you need any help with it?

lolmaus:
@herzzanu That's a question to addon maintainers. 😬

herzzanu:
Thanks @lolmaus 

@cibernox this one looks promising. How can we help to move it forward? 🙏

herzzanu:
I have a branch here based on the last version and inspired by the changes in this PR https://github.com/EmberExperts/ember-basic-dropdown

apellerano-pw:
> I have a branch here based on the last version and inspired by the changes in this PR https://github.com/EmberExperts/ember-basic-dropdown/tree/shadow-dom

I was able to upgrade to this newer fork with no issues. We've been using the kaliber5 fork for years

apellerano-pw:
Oh I found an issue in the EmberExperts fork. Inside `component:basic-dropdown-content`, it still uses the old way of registering the `triggerElement` so that `handleRootMouseDown` expects it to exist at `this.args.triggerElement`. However, the newer version of ember-basic-dropdown no longer uses a component arg. Instead it looks up the trigger element just in time using a data attribute. (Which would need to be shadow DOM aware.)

I'll see if I can get an example commit tomorrow

apellerano-pw:
This commit on top of the EmberExperts fork addresses the issue when clicking the trigger to close the dropdown content.

I scanned for any other dangling `document.bleh` element lookups and found 2 more, made sure they are trying out `owner.rootElement` as well.

https://github.com/ProsperWorks/ember-basic-dropdown/commit/21d4c7dbe72b2b43e8ba4b999870a30e1ecd6fd6

mkszepp:
I would like to close/merge this PR... can sombody bring this PR up to date? Unfortunately there is not anymore mergable...

apellerano-pw:
I've made my fixes to the EmberExpert fork available here https://github.com/EmberExperts/ember-basic-dropdown/pull/1

Maybe @herzzanu can incorporate my changes and open a new PR here?

herzzanu:
Thanks for your contribution @apellerano-pw 🙏  I've merged that PR and created a [PR here](https://github.com/cibernox/ember-basic-dropdown/pull/789) to incorporate the changes back into this repo. We need to fix the conflicts 🙈 and should be good to go.

apellerano-pw:
I see, the changes don't apply cleanly because the addon has been migrated to v2 format. I'm not enough of a git plumber to know the easy way to approach this. I'll try to re-apply our commits onto HEAD and open a PR

mkszepp:
closed by #791

Issue #615, opened 2021-05-20
Transition animation doesn't reset after closing/reopening


kevinkucharczyk:
This can be seen even in the docs: https://ember-basic-dropdown.com/docs/animations/

The `transitioning--in` etc classes don't get reset after the first render. When the dropdown content first renders, it animates correctly, but on subsequent renders there is no animation - it appears instantly.

Here'a a screencast:
![2021-05-20 11-27-10 2021-05-20 11_27_30](https://user-images.githubusercontent.com/3127419/118905120-7b33f880-b95e-11eb-9b73-88267164ee79.gif)

You can see how on the first click in smoothly animates in, but after that it doesn't.

There also doesn't seem to be a transitioning-out animation, even though it's defined in the CSS.

denzo:
Yes, I am experiencing the same in `ember-power-select` as it uses this addon under the hood.

Can't find `dropdown--transitioning-out` in the codebase at all also.

mkszepp:
closing this old issue... looks like it was fixed already (at least in our docs it works)

PR #623, opened 2021-06-19
Fix for https://github.com/cibernox/ember-basic-dropdown/issues/615

ducharmemp:
Fixes https://github.com/cibernox/ember-basic-dropdown/issues/615 🤞 

I noticed that as of https://github.com/cibernox/ember-basic-dropdown/commit/1819c011578dbd23f969b08501ca1685d01dcdf8#diff-73297f8c71645544a2a6826e5c3146aff7cb2333d9c17fe968258e5f14dc4a79R164 the animationClass was set to `transitionedInClass` when animating out. I set this back to the original value, and also reordered the calls to preemptively set the state on transitioning out, regardless if the parent element exists. I _think_ this is correct because no matter what the reason (barring animation disabled), on a transition out we want to be able to render with the correct value when the drop down enters the dom again.

### DIFF
--- a/addon/components/basic-dropdown-content.ts
+++ b/addon/components/basic-dropdown-content.ts
@@ -125,6 +125,7 @@ export default class BasicDropdownContent extends Component<Args> {
   @action
   animateOut(dropdownElement: Element): void {
     if (!this.animationEnabled) return;
+    this.animationClass = this.transitioningInClass;
     let parentElement = dropdownElement.parentElement;
     if (parentElement === null) return;
     if (this.args.renderInPlace) {
@@ -136,7 +137,6 @@ export default class BasicDropdownContent extends Component<Args> {
     clone.classList.remove(...this.transitioningInClass.split(' '));
     clone.classList.add(...this.transitioningOutClass.split(' '));
     parentElement.appendChild(clone);
-    this.animationClass = this.transitionedInClass;
     waitForAnimations(clone, function() {
       (parentElement as HTMLElement).removeChild(clone);
     });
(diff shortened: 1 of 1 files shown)


ducharmemp:
Note that on the preview site, animation out is still semi-broken. Not sure exactly _why_ but I have a hunch that `will-destroy` is getting called in the template after the parent node has been removed from the DOM, so we're exiting the function early. Local testing confirms this is the case but I'm not sure how or why considering the docs say it'll be called _before_ all that happens.

https://github.com/emberjs/ember-render-modifiers/blob/master/addon/modifiers/will-destroy.js#L4-L5

addon/components/basic-dropdown-content.ts
```diff
@@ -125,18 +125,14 @@ export default class BasicDropdownContent extends Component<Args> {
   @action
   animateOut(dropdownElement: Element): void {
     if (!this.animationEnabled) return;
-    let parentElement = dropdownElement.parentElement;
-    if (parentElement === null) return;
-    if (this.args.renderInPlace) {
-      parentElement = parentElement.parentElement
-    }
+    this.animationClass = this.transitioningInClass;
+    let parentElement = this.destinationElement;
```
ducharmemp, line 129: Not sure if this is the right fix, but what I was seeing is that the parentElement was always null on `will-destroy` calls. However, since we want to animate at the destination location I think this is right? Not sure 🤷‍♂️

ducharmemp:
> Seems like this fix isn’t working if you are using `renderInPlace` but not sure? 

You might be right, I can't remember if I tested with render in place. I'll give it a look this evening

dardalan:
Hey @ducharmemp! Any updates on this issue?

dardalan:
@ducharmemp @emilkarl imho, it should be just enough to revert this 
https://github.com/cibernox/ember-basic-dropdown/commit/1819c011578dbd23f969b08501ca1685d01dcdf8#diff-73297f8c71645544a2a6826e5c3146aff7cb2333d9c17fe968258e5f14dc4a79R164

```
this.animationClass = this.transitionedInClass;
change to 
this.animationClass = this.transitioningInClass;
```

Issue #589, opened 2020-12-07
Implement Trigger as a class element modifier?


Techn1x:
Hi! I have had some trouble using ebd, regarding the trigger element, issues mostly due to the fact the trigger and the content of the trigger, are separate elements. For example;
```hbs
<dd.Trigger>
  <MyUiButton class="some-class" @size="lg" />
</dd.Trigger>
```

- focus/tabbing through elements requires 2 tabs, since they're separate tabable elements
- potential layout weirdness - if the ebd trigger element is in a flex, it'll expand, but the button component is an inline-block by default, so it ends up being smaller than the trigger, which may or may not be immediately obvious to the dev that a click to the side of the button will open the dropdown.

I figured these could be solved by implementing the dropdown trigger as an element modifier, and then any component or element can be used _as the trigger itself_. Something along the lines of;
```hbs
<BasicDropdown ... |dd|>
  <MyUiButton class="some-class" @size="lg" {{dropdown-trigger dropdown=dd}} />
  <dd.Content> ... </dd.Content>
</BasicDropdown>
```

EDIT: I've made a new PR over at: https://github.com/cibernox/ember-basic-dropdown/pull/626

Are the maintainers open to this idea? I don't think I would be the one to implement it fully, there's a lot of little bits and pieces and history I don't really understand - hoping to inspire someone else to run with the idea.

Techn1x:
Any thoughts on this? I could potentially do the work if there was some support behind the idea

burritoIand:
@Techn1x I believe this is already possible (to use a custom element as the trigger).

https://ember-basic-dropdown.com/cookbook/no-trigger

```
<BasicDropdown as |dd|>
  <UiButton @dd={{dd}}/>
    Click me
  </UiButton>

  <dd.Content>
    hello world
  </dd.Content>
</BasicDropdown>
```

```
<button
  {{on "click" @dd.actions.toggle}}
>
  {{yield}}
</button>
```

Techn1x:
Oh. wow thankyou @burritoIand - I think that might actually be exactly what I was after!

burritoIand:
@Techn1x np! lmk if you have any questions // @guac on sqwok.im too

Techn1x:
[...]

PR #588, opened 2020-12-07
Implement Trigger as a class element modifier (WIP)

Techn1x:
_This is a WIP / draft and is not intended to be merged, just here for feedback_

Moves dropdown trigger component logic (most of it) into an ember class modifier.

The existing usage remains intact, but allows for _any_ component or element to be made into a trigger with relative ease.

existing usage
```hbs
<BasicDropdown ... |dd|>
  <dd.Trigger> my trigger </dd.Trigger>
  <dd.Content> ... </dd.Content>
</BasicDropdown>
```

additional usage with a modifier
```hbs
<BasicDropdown ... |dd|>
  <MyComponent {{dropdown-trigger dropdown=dd}} />
  <dd.Content> ... </dd.Content>
</BasicDropdown>
```

Proof that it does indeed work as a modifier
Peek 2020-12-07 14-04

PR #626, opened 2021-07-14
Implement Trigger as a class element modifier

Techn1x:
Closes https://github.com/cibernox/ember-basic-dropdown/issues/589
Supersedes https://github.com/cibernox/ember-basic-dropdown/pull/588

Moves all of the dropdown trigger component logic into an ember class modifier.

The existing usage remains intact, but allows for any component or element to be made into a trigger with relative ease, keeping all the handy a11y and bindings.

usage with a modifier;
```hbs
<BasicDropdown ... |dd|>
  <MyComponent {{basic-dropdown-trigger dropdown=dd}} />
  <dd.Content> ... </dd.Content>
</BasicDropdown>
```

I also gave a bit of TLC to some of the logic as I was moving it across into the modifier. I'll comment on the noteworthy changes.

Peek 2021-07-15 01-17


addon/modifiers/basic-dropdown-trigger.ts
```diff
@@ -0,0 +1,137 @@
...
+
+    this.element.removeEventListener('click', this.handleMouseEvent);
+    this.element.removeEventListener('mousedown', this.handleMouseEvent);
+    this.element.removeEventListener('keydown', this.handleKeyDown);
+    this.element.removeEventListener('touchstart', this.handleTouchStart);
+    this.element.removeEventListener('touchend', this.handleTouchEnd);
+  }
+
+  @action
+  handleMouseEvent(e: MouseEvent): void {
```
Techn1x, line 57: The `handleMouseDown` and `handleClick` actions were very similar, so I consolidated them into this action

addon/modifiers/basic-dropdown-trigger.ts
```diff
@@ -0,0 +1,137 @@
...
+  named: {
+    dropdown: Dropdown
+    eventType?: 'click' | 'mousedown'
+    stopPropagation?: boolean
+  }
+}
+
+export default class DropdownTriggerModifier extends Modifier<Args> {
+  @tracked private toggleIsBeingHandledByTouchEvents: boolean = false
+  @tracked private touchMoveEvent?: TouchEvent
```
Techn1x, line 18: These weren't tracked before, I feel like they should have `@tracked`

addon/components/basic-dropdown-trigger.ts
```diff
@@ -32,121 +27,4 @@ export default class BasicDropdownTrigger extends Component<Args> {
...
-  @action
-  handleMouseDown(e: MouseEvent): void {
-    if (this.args.dropdown.disabled) {
-      return;
-    }
-    if (this.args.eventType !== 'mousedown' || e.button !== 0) return;
-    if (this.args.stopPropagation) {
-      e.stopPropagation();
-    }
-    this._stopTextSelectionUntilMouseup();
```
Techn1x, line 45: I removed this function. All it seemed to do was add & remove `ember-basic-dropdown-text-select-disabled` class name to the document body, a class name not referenced anywhere in this addon. I think it's just something from a long time ago

addon/templates/components/basic-dropdown-trigger.hbs
```diff
@@ -4,21 +4,13 @@
     {{if @renderInPlace " ember-basic-dropdown-trigger--in-place"}}
     {{if @hPosition (concat " ember-basic-dropdown-trigger--" @hPosition)}}
     {{if @vPosition (concat " ember-basic-dropdown-trigger--" @vPosition)}}
-
     {{@defaultClass}}"
```
Techn1x, line 7: Could use some feedback on this one. I've opted not to include the above class names in the modifier. I felt that if a user is implementing a custom trigger using the modifier, they have direct access to the element so they can apply any class names they want to apply

Let me know if I should include them. I could do that just to be safe.

tests/integration/components/trigger-test.js
```diff
@@ -57,17 +55,17 @@ module('Integration | Component | basic-dropdown-trigger', function (hooks) {
...
-    this.dropdown = { uniqueId: 123 };
-    await render(hbs`
-      <BasicDropdownTrigger tabindex={{false}} @dropdown={{this.dropdown}}>Click me</BasicDropdownTrigger>
-    `);
+  // test('If it receives a tabindex={{false}}, it removes the tabindex', async function (assert) {
+  //   assert.expect(1);
+  //   this.dropdown = { uniqueId: 123 };
+  //   await render(hbs`
+  //     <BasicDropdownTrigger tabindex={{false}} @dropdown={{this.dropdown}}>Click me</BasicDropdownTrigger>
+  //   `);
```
Techn1x, line 63: Could use some feedback on this one - This is the only test I haven't been able to get passing.

When `tabindex={{false}}`, that is presented as `null` once we're inside the modifier (null indicates an attribute is not set). This is also what the modifier uses to determine if it should apply tabindex="0" (and auto-manage it when disabled)

In other words, there's no way for the modifier to know the difference between
`<BasicDropdownTrigger tabindex={{false}} ... />`
and
`<BasicDropdownTrigger ... />`

Any ideas?

addon/components/basic-dropdown-trigger.ts
Techn1x: (I can add it back if that causes problems)

Techn1x:
Take 2 @cibernox - other than the commented out test, this is hopefully be good to go. All other tests passing locally.

addon/modifiers/basic-dropdown-trigger.ts
```diff
@@ -0,0 +1,138 @@
...
+      this.element.setAttribute('aria-disabled', 'true');
+    } else {
+      // only set to 0 if tab index is not a custom value
+      if (this.element.getAttribute('tabindex') === null) this.element.setAttribute('tabindex', '0');
+      this.element.setAttribute('aria-disabled', 'false');
+    }
+  }
+
+  willRemove() {
+    if (typeof document !== 'undefined') document.removeEventListener('touchmove', this._touchMoveHandler);
```
Techn1x, line 48: not sure why we have this `typeof document !== 'undefined'` check when optional chaining should suffice `document?.removeEventListener('touchmove', this._touchMoveHandler);`

I figured it might be a browser compatiblity thing?

addon/templates/components/basic-dropdown-trigger.hbs
```diff
@@ -4,21 +4,13 @@
...
   ...attributes
-  {{will-destroy this.removeGlobalHandlers}}
-  {{on "mousedown" this.handleMouseDown}}
-  {{on "click" this.handleClick}}
-  {{on "keydown" this.handleKeyDown}}
-  {{on "touchstart" this.handleTouchStart}}
-  {{on "touchend" this.handleTouchEnd}}
   {{! V1 compatibility - See #498 }}
   {{on "keydown" (fn (or @onKeyDown this.noop) @dropdown)}}
   {{on "mousedown" (fn (or @onMouseDown this.noop) @dropdown)}}
```
Techn1x, line 16: didn't feel the need to include the legacy V1 events stuff in the modifier. Anyone using the modifier shouldn't be using the old events args

tests/integration/components/trigger-test.js
```diff
@@ -127,10 +125,10 @@ module('Integration | Component | basic-dropdown-trigger', function (hooks) {
     assert
       .dom('.ember-basic-dropdown-trigger')
       .hasAttribute('aria-disabled', 'true', 'It is marked as disabled');
-    run(() => this.set('dropdown.disabled', false));
+    this.set('dropdown', { ...this.dropdown, disabled: false });
     assert
       .dom('.ember-basic-dropdown-trigger')
-      .doesNotHaveAttribute('aria-disabled', 'It is NOT marked as disabled');
+      .hasAttribute('aria-disabled', 'false', 'It is NOT marked as disabled');
```
Techn1x, line 131: Opted to explicitly set `aria-disabled` to "true" or "false", rather than removing the attribute for false. `aria-disabled` works differently to the standard `disabled` attribute which only cares about its presence (it will still disable the element when "false", if that makes sense)

Techn1x:
Hi @cibernox - I've recently encountered the need for this to be a modifier again and sorely wishing this work was part of main.

If I were to update this work to deal with merge conflicts & use the new modifier API for the trigger, would you consider merging it?

Techn1x:
oh sorry, I just re-read your comment on the prior/closed PR
> Can you rebase this? I love the idea, I've been wanting to do this since before element modifiers existed.

I'll get to work and I'll be more pro-active in getting this merged this time.

Techn1x:
OK, this is ready to rock @cibernox - let me know if I should ping someone else if you're busy :heart: 

I really think a modifier is an excellent way to go here for this addon, and we could feasibly even deprecate the trigger _component_ itself in the future, in favour of direct use of the modifier. It's basically a shell now, only applying class names and legacy event handlers. 

In theory, there are no breaking changes here. So a minor version bump should do it.

Techn1x:
friendly ping - @cibernox , noticed you merged some PRs a few hours ago (which have caused more merge conflicts here). I would really appreciate a response from you (of any kind). This will be the 4th or 5th time I've had to update this PR / rebase

addon/components/basic-dropdown-trigger.ts
cibernox: The reason for this class is so we can add `user-select: none;` to the document's body temporarily. This is to allow someone to mousedown on the trigger of a select, and without lifting its finger, move to the chosen option and lift then finder from the mouse. Since the dropdown is not physically (in the DOM) adjacent to the trigger, that flow tended to highlight the entire page.

I think we need it back.

cibernox:
Oh, I've been disconnected from open source for a few months, with just 15 minutes here and there to check stuff. I added a couple comments, but now that I've started checking this i wont let it stale again.

I saw that Ember 3.13 is failing. If this can't be made to work that far back I'm cool bumping the minimum required version of Ember to something newer, I'll probably bump the version number anyway after this.

Techn1x:
> Oh, I've been disconnected from open source for a few months, with just 15 minutes here and there to check stuff. I added a couple comments, but now that I've started checking this I wont let it stale again.

All good - I've been the same, which is why there's 6 months between each of my stints working on this

> I saw that Ember 3.13 is failing. If this can't be made to work that far back I'm cool bumping the minimum required version of Ember to something newer, I'll probably bump the version number anyway after this.

Ah! I don't think I was able to see those checks before. That'll be due to ember-modifier. Looking at their compatibility, it seems it is https://github.com/ember-modifier/ember-modifier#compatibility
> Ember.js v3.24 or above

addon/components/basic-dropdown-trigger.ts
Techn1x: > The reason for this class is so we can add user-select: none; to the document's body temporarily

Good to know the reason - but as far as I can tell it's not doing that. There's no CSS for the `ember-basic-dropdown-text-select-disabled` class, and there's nothing that applies `user-select` style anywhere in this addon
https://github.com/cibernox/ember-basic-dropdown/search?q=ember-basic-dropdown-text-select-disabled&type=code
https://github.com/cibernox/ember-basic-dropdown/search?q=user-select&type=code

Or was this class intended for users of this addon to implement styles for?

addon/modifiers/basic-dropdown-trigger.ts
```diff
@@ -0,0 +1,157 @@
...
+  stopPropagation?: boolean;
+  [named: string]: unknown;
+}
+type PositionalArgs = unknown[];
+interface Args {
+  positional: PositionalArgs;
+  named: NamedArgs;
+};
+
+export default class DropdownTriggerModifier extends Modifier<Args> {
```
Techn1x, line 19: One thing I've learned only in the last couple days - the new Modifier API has a Signature type that we can leverage here for modifier args.
https://github.com/ember-modifier/ember-modifier#the-signature-type

It won't functionally affect this addon, and I'm a little busy over the Easter weekend, so I will probably do it as a follow up PR

cibernox:
@Techn1x nope, go for it and make it 2.24+ only. 3.13 was released 2 years and a half ago, it's totally understandable that at some point things would stop working on it. I'll just publish it with a major version bump.

addon/components/basic-dropdown-trigger.ts
cibernox: If memory serves, I never included default styles for this component but some other component i have based on this one (ember-power-datepicker maybe? It could be another) had styles targeting that class.

addon/modifiers/basic-dropdown-trigger.ts
```diff
@@ -0,0 +1,157 @@
...
+    element.addEventListener('mousedown', this.handleMouseEvent);
+    element.addEventListener('keydown', this.handleKeyDown);
+    element.addEventListener('touchstart', this.handleTouchStart);
+    element.addEventListener('touchend', this.handleTouchEnd);
+  }
+
+  update(element: HTMLElement, _positional: PositionalArgs, named: NamedArgs) {
+    const { dropdown } = named
+
+    element.setAttribute('data-ebd-id', `${dropdown?.uniqueId}-trigger`);
```
cibernox, line 56: The only thing that concerns me is that updates set all those attributes even if they haven't changed. 
Do you know if modifiers have a way of only updating what has changed? 

Probably not a big deal, it's going to be performant enough, but I wondered.

addon/modifiers/basic-dropdown-trigger.ts
cibernox: If tests pass, I'm happy with this being one.

addon/components/basic-dropdown-trigger.ts
Techn1x: Ah ok, will add back in. Looks like a lot of people definitely copied some styles from somewhere for it
https://github.com/search?l=CSS&q=ember-basic-dropdown-text-select-disabled&type=Code

addon/modifiers/basic-dropdown-trigger.ts
Techn1x: We can set a tracked property on the class, let's say `uniqueId`, then only set those attrs if that id changed
`if (dropdown?.uniqueId !== this.uniqueId) this.setUniqueAttributes(dropdown?.uniqueId)`

I felt the simplicity was more important

addon/components/basic-dropdown-trigger.ts
Techn1x: Added this back in to the component, rather than the modifier - was intending to keep the modifier without much frills or any external side effects (like adding classes to body element). Seems to work great

addon/modifiers/basic-dropdown-trigger.ts
Techn1x: implemented this

addon/modifiers/basic-dropdown-trigger.ts
cibernox: I'd say that you don't have to worry. The trigger is a single component, so performance is not so critical. I would be more concerned if this was ember-power-select and this was each of the possibly hundreds of options. Keep it this simple.

cibernox:
Ok, CI is green. If you think this is ready, I can merge it and release it as a beta tomorrow

Techn1x:
Good to go. Thanks heaps

cibernox:
Merged! Thank you so much for this outstanding work. I think it's going to make so much easier to create custom triggers

cibernox:
I just published 6.0.0

PR #630, opened 2021-10-07
Update render modifiers

jgadbois:
v1.0.2 threw a lot of deprecation warnings on newer Ember versions.

Tests still pass after this update


josemarluedke:
I'm also looking for this change to be merged. 

But I think this PR needs to update the package-lock.json as well.

jgadbois:
Good call - updated

PR #637, opened 2021-12-01
Avoid deprecation for `import { assign } from '@ember/polyfills'`

mydea:
Use `Object.assign` instead of `Ember.assign`, as that is deprecated here: https://github.com/emberjs/rfcs/blob/master/text/0750-deprecate-ember-assign.md

This basically means that support for IE11 is dropped. I guess you could still run this with IE11 by providing a global polyfill yourself for `Object.assign`.

### DIFF
--- a/addon/components/basic-dropdown.ts
+++ b/addon/components/basic-dropdown.ts
@@ -295,8 +295,7 @@ export default class BasicDropdown extends Component<Args> {
       }
     }
     for (let prop in positions.style) {
-      // Array.includes is not available for IE11
-      if (IGNORED_STYLES.indexOf(prop) === -1) {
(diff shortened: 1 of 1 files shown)


cibernox:
IMHO, I think we can just get rid of the polyfill altogether. It's almost 2022, I think we're good. 
I'll release a new major version to signal that it removes compatibility with IE never look back.

Do you want to do the honors?

mydea:
Sure thing! I'll adjust the PR then, sounds good to me ;)

mydea:
I adjusted the PR to stop using @ember/polyfills!

cibernox:
Merged! Thanks!

PR #628, opened 2021-08-19
Fixed state not being reset to correct value when transitioning out

dardalan:
This is already opened pr on this Issue cibernox#623
but i am not sure we should make any changes on this code [ember-basic-dropdown](https://github.com/cibernox/ember-basic-dropdown/pull/623/files#diff-5eb8bde405ea618cba05ea05b9898dc60ff96ac459a953c279fbb24f39a46f81L128-L132), cause on production code - it works good, the only problem with tests, so i made a workaround for testing purpose. Any idea how to fix this issue better, very welcome.

The transitioning--in etc classes don't get reset after the first render. When the dropdown content first renders, it animates correctly, but on subsequent renders there is no animation - cause state not being reset to correct value.

Looks like it was broken here in this commit, so idea to revert back and cover with tests:
cibernox@1819c01#diff-73297f8c71645544a2a6826e5c3146aff7cb2333d9c17fe968258e5f14dc4a79R164

PR #639, opened 2022-01-05
Make sure the content parent element is set before animating out

gwak:
This is a follow-up to the PRs #623 and #628. The fix on the animation flow worked well when the dropdown is rendered in place. Not so well when not because the parent element is lost when the content element is closed.

In the `animateOut` function, there's an early exit when no `dropdownElement.parentElement` is null. So before the check I just try to get the parentElement through the `destinationElement`.

### DIFF
--- a/addon/components/basic-dropdown-content.ts
+++ b/addon/components/basic-dropdown-content.ts
@@ -54,7 +54,7 @@ export default class BasicDropdownContent extends Component<Args> {
   private handleRootMouseDown?: RootMouseDownHandler;
   private scrollableAncestors: Element[] = [];
   private mutationObserver?: MutationObserver;
-  @tracked animationClass = this.transitioningInClass
+  @tracked animationClass = this.transitioningInClass;
 
   get destinationElement(): Element | null {
     return document.getElementById(this.args.destination);
@@ -149,7 +149,8 @@ export default class BasicDropdownContent extends Component<Args> {
   @action
   animateOut(dropdownElement: Element): void {
     if (!this.animationEnabled) return;
-    let parentElement = dropdownElement.parentElement;
+    let parentElement =
+      dropdownElement.parentElement ?? this.destinationElement;
     if (parentElement === null) return;
     if (this.args.renderInPlace) {
       parentElement = parentElement.parentElement;
@@ -161,7 +162,7 @@ export default class BasicDropdownContent extends Component<Args> {
     clone.classList.add(...this.transitioningOutClass.split(' '));
     parentElement.appendChild(clone);
     this.animationClass = this.transitioningInClass;
-    waitForAnimations(clone, function() {
+    waitForAnimations(clone, function () {
       (parentElement as HTMLElement).removeChild(clone);
     });
   }


gwak:
Hello, how can I help to make this PR merged.

cibernox:
Sorry, I missed this PR completely

cibernox (approved):

cibernox:
Merged! I'm going to cut a new release.

cibernox:
4.0.3 contains this change. Thanks again 🙏

gwak:
@cibernox Thank you!

addon/components/basic-dropdown-content.ts
```diff
@@ -149,7 +149,8 @@ export default class BasicDropdownContent extends Component<Args> {
   @action
   animateOut(dropdownElement: Element): void {
     if (!this.animationEnabled) return;
-    let parentElement = dropdownElement.parentElement;
+    let parentElement =
+      dropdownElement.parentElement ?? this.destinationElement;
```
hoIIer, line 153: this is causing a weird double animate out in my app

gwak: Hello, thanks for the feedback @hoIIer, I can have a look, but could you please be more specific on your context, like:

- Browser + OS (Does the [animation on the docs](https://ember-basic-dropdown.com/docs/animations) behaves normally on your browser ?)
- Animation css you're using
- renderInPlace or not
- ...

cibernox: @hoIIer Also interested, It didn't cause any problem on mine.

hoIIer: - happens in both chrome/firefox latest in osx
- using custom calculated position that sets css animation dynamically

```
animation: "popoutFadeInSlideLeft 0.1s ease-in-out"
right: 709.5
top: 433.1796875
```
```
  @keyframes popoutFadeInSlideLeft {
    from {
      opacity: 0;
      transform: translateX(-15px);
    }
    to {
      opacity: 1;
      transform: translateX(0px);
    }
  }
```
When I remove the `style.animation = ...` the popout looks normal

```
<BasicDropdown
  @calculatePosition={{this.calcPosition}}
  as |dd|
>
  {{yield (hash
    dd=dd
    trigger=(
      component 'popout-menu/trigger'
      componentStyle=(get-style this)
      disableMenu=@disableMenu
      dropdown=dd
      name=@name
      visible=@visible
    )
    menu=(
      component 'popout-menu/menu'
      componentStyle=(get-style this)
      dropdown=dd
      isDesktop=this.media.isDesktop
      onClickOverlay=this.onClickOverlay
      theme=@theme
    )
    componentStyle=(get-style this)
  )}}
</BasicDropdown>
```

hoIIer: I extended the animation duration to 10 seconds and noticed when closing the dropdown, the content menu is immediately reset to it's starting position, but it still has the animation value set so it does a 2nd animation in before closing.

hoIIer: Confirmed that in `4.0.2` the content menu does not reset to original position after clicking out to close the dropdown, the `<div id="ember-basic-dropdown-wormhole"></div>` just becomes empty again.

In `4.0.3` when clicking out to close the dropdown, the content menu appears again inside the parent container and with a `--clone` css class, leading to the 2nd animation before it's removed entirely.

hoIIer: Can revisit later today, perhaps I need to adjust how I'm setting the animation? In this case I have to set it dynamically within the calcPosition function because my contextual dropdown is used in different contexts.

PR #643, opened 2022-01-26
Update @embroider packages to 1.x

mydea:
This makes it much easier to handle dependencies between addons via `^` versions.

In the 0.50.1 update for @embroider/macros, there was also a (somewhat hidden) change introduced that changed the behavior of `dependencySatisfies` to require being defined as a peer dependency (see https://github.com/embroider-build/embroider/pull/1070). IMHO that is not really worth it for this use case here, and we can just as well use a try-catch block in that case. Happy to change it back and add @ember/test-helpers as an optional peer dependency, if you prefer that.


cibernox:
Thanks!

PR #644, opened 2022-01-26
Improve CI workflow

mydea:
Currently, all CI steps run `npm test`, which in turn will run `npm test:ember-compatibility`, which will run the tests for all ember-try scenarios.

This PR streamlines this to:

* `tests` which runs the regular ember tests
* `floating-dependencies` which has been slightly adjusted to run `npm install --lock-file=false` to ensure it is truly floating
* `try-scenarios` which actually runs the ember-try scenarios in a matrix

I also removed the unused node matrixes (as there was now everywhere just a single matrix item `14`). Both floating-dependencies & try-scenarios are now run only after tests completed. Finally, the concurrency setting ensures that a CI run is cancelled if a new state is pushed to the same branch while CI is running.

The visualisation shows the change nicely:

Original:
Screenshot from 2022-01-26 13-53-22

New:
Screenshot from 2022-01-26 13-53-10

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -12,10 +12,14 @@ on:
     branches:
       - master
   schedule:
-    - cron: '0 3 * * 0' # every Sunday at 3am
+    - cron: "0 3 * * 0" # every Sunday at 3am
 
 env:
   CI: true
+
+concurrency:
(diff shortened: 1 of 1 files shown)


cibernox:
This is indeed so much better. Thanks!

PR #650, opened 2022-01-31
Fix misleading documentation for destination

ef4:
The destination option doesn't accept a selector, it only accepts an id.

There is [an open PR](https://github.com/cibernox/ember-basic-dropdown/issues/406) about changing that, but it never landed.


cibernox:
thanks!

PR #653, opened 2022-02-07
Fix docs mobile layout

gwak:
This PR fixes the docs styles on mobile layout. 

1. The width of the content wrapper divs was fixed to 1024px with a max-width of 1024px. So I switched the width to be 100%. 

https://user-images.githubusercontent.com/176766/152803909-444293cc-5bb3-4e93-83a5-3e93ef5241f6.mov

2. I also added an `overflow-x` to the header links on mobile in case it overflows
3. In the `docs` page, I made the `groupedSections` the source of truth for both navigations (main nav and the native select on mobile)
4. It was missing the animations page on mobile

### DIFF
--- a/tests/dummy/app/controllers/public-pages/docs.js
+++ b/tests/dummy/app/controllers/public-pages/docs.js
@@ -4,6 +4,7 @@ import Controller from '@ember/controller';
 
 const groupedSections = [
   {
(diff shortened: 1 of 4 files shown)


cibernox:
Thanks!

PR #658, opened 2022-03-09
Only set animationClass on dropdown-content when animations are enabled

nickschot:
This fixes an issue where the dropdown content would be invisible in tests due to the transitioning in class always being set. Broke visual regression testing etc.


nickschot:
🤦 oh this was added here as a fix for app behaviour: https://github.com/cibernox/ember-basic-dropdown/pull/628

Double checking... but the current behaviour is also not great as it basically breaks tests/percy in the way I described

PR #661, opened 2022-04-08
Co-locate component templates

ef4:
This PR co-locates the component templates with their JS files.

This is an important change because without it users cannot easily adopt [first-class component templates](https://github.com/emberjs/rfcs/blob/master/text/0779-first-class-component-templates.md).

Co-located templates stay bound to the corresponding class even when you pass the class through arbitrary Javascript. Non-co-located templates don't.

Component template co-location is supported in all the ember versions this addon supports. So this should not be a breaking change. Any users who were overriding these templates with their own implementations *would* break, but those users are necessarily relying on private API (the API between each component JS and its own template).


cibernox:
I'll merge this and release it as a new major version then.

Thanks!

PR #675, opened 2022-10-03
find dropdown root using trigger.parentElement

### LINKED ISSUE #583: Animation only plays for initial open of dropdown
This issue is present on the documentation site on the "Animations" page, tested on latest version of Chrome (Mac and Windows) as well as Mac Safari and Mac Firefox:

Aierie:
This should fix #583

`will-destroy` can run at the [same tick as dropdown content removal](https://github.com/pzuraq/emberjs-rfcs/blob/render-element-modifiers/text/0000-render-element-modifiers.md#will-destroy). If it is indeed removed from the DOM, we would not be able to find the container by starting from the dropdown content, so this PR uses the trigger to find the container instead. 

It seems difficult to test this - I've tried using paused animations (but this doesn't play nice with `waitForAnimations` which only waits for a running animation), however I'd be happy to implement suggestions.

### DIFF
--- a/addon/components/basic-dropdown-content.ts
+++ b/addon/components/basic-dropdown-content.ts
@@ -150,12 +150,10 @@ export default class BasicDropdownContent extends Component<Args> {
   animateOut(dropdownElement: Element): void {
     if (!this.animationEnabled) return;
     let parentElement =
-      dropdownElement.parentElement ?? this.destinationElement;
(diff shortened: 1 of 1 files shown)


Aierie:
Closing this for now - realised that the named block isn't strict about where you render your dropdown content, so this may mess up behaviour of some dropdowns.

PR #687, opened 2022-12-10
fix: compile with exactOptionalPropertyTypes and noImplicitOverride

bwbuchanan:
TypeScript consumers of this addon that build with the TS options `exactOptionalPropertyTypes` or `noImplicitOverride` currently fail to compile.

This PR fixes the issue by explicitly marking overridden functions as `override` and adding an explicit `| undefined` for optional interface & class members that can be assigned as undefined.


bwbuchanan:
Perhaps this addon's `tsconfig.json` should also specify those options, to avoid future regressions?

In general, I think that addons should be built with the strictest possible constraints on their types, to avoid breaking downstream consumers that want to use stricter `compilerOptions`.

cibernox:
I also think that would be ideal. Compile this as strictly as possible

cibernox:
Merged and published!

Issue #550, opened 2020-05-04
animationEnabled arg ignored by BasicDropdownContent component


tzellman:
This is occurring in `v3.0.1`, possibly others, but the only version I am locally using/testing.

When dealing with an issue related to #540 I was hoping to get around the error by disabling animation.

According to the published docs, `BasicDropdownContent` supports an `animationEnabled` arg, however it appears that the code is not using the arg in the `animationEnabled` getter.

Is this intentional and the docs are outdated (i.e. 2.x)? Or should a user be able to specify:

```hbs
<dd.Content @animationEnabled={{false}} >
```

I can follow-up this issue with a PR. ❤️

ArnaudWeyts:
Also running into this issue, it seems like the animation is implemented by cloning and appending the dropdown element to the DOM when it's being destroyed, then applying the transition class to the cloned element. This causes web components inside of the dropdown list to be re-rendered using empty arguments and erroring. Being able to disable the animation would already help, but it seems like it's only disabled when in testing mode: https://github.com/cibernox/ember-basic-dropdown/blob/77d04df8fa2cafb415b528c3bcf2f90b0a5861b0/addon/components/basic-dropdown-content.ts#L63-L65

PR #700, opened 2023-02-21
Allow passing @animationEnabled argument to BasicDropdownContent

ArnaudWeyts:
Fixes https://github.com/cibernox/ember-basic-dropdown/issues/550

This argument is documented in both ember-basic-dropdown and ember-power-select, but never actually used or passed along. I'll also create a PR in ember-power-select to pass the argument along.


cibernox (approved):

cibernox:
Thanks!

PR #710, opened 2023-07-28
Update ember-element-helper to v0.7.1

mkszepp:
There was released a new version of [ember-element-helper](https://github.com/tildeio/ember-element-helper/blob/v0.7.1/ember-element-helper/CHANGELOG.md). No braking changes, only added support for `ember-source` v5.


mkszepp:
@cibernox can we get a release with this changes?

PR #714, opened 2023-10-22
Drop `@ember/render-modifiers`

gilest:
Replaces the no-longer-recommended [render lifecycle modifiers](https://github.com/emberjs/ember-render-modifiers) with local function modifiers.

`@ember/render-modifiers` were meant as a transitional tool from the pre-Octane era, and not long-term usage.

Modifier code is directly ported from the setup/teardown hooks with no changes except what I've pointed out in code comments.

Recommend reviewing commit-by-commit.

### DIFF
--- a/addon/components/basic-dropdown-content.ts
+++ b/addon/components/basic-dropdown-content.ts
@@ -11,6 +11,7 @@ import {
 import hasMoved from '../utils/has-moved';
 import { Dropdown } from './basic-dropdown';
 import { isTesting } from '@embroider/macros';
+import { modifier } from 'ember-modifier';
 
 interface Args {
   animationEnabled?: boolean;
@@ -66,7 +67,7 @@ export default class BasicDropdownContent extends Component<Args> {
 
     return animationEnabledArg && !isTesting();
   }
-  
+
   get positionStyles(): Record<string, string> {
     const style: Record<string, string> = {};
     if (this.args.top !== undefined) {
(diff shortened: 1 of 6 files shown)


addon/components/basic-dropdown-content.ts
```diff
@@ -137,61 +137,65 @@ export default class BasicDropdownContent extends Component<Args> {
...
+        document.removeEventListener(
+          'touchend',
+          this.handleRootMouseDown as RootMouseDownHandler,
+          true
+        );
+      }
     }
-  }
+  // @ts-ignore
+  }, { eager: false });
```
gilest, line 161: All `modifier()` invocations must include `{ eager: false }` since this package supports `ember-modifier` v3 and v4.

For apps using v3.2.x `{ eager: false }` will ensure the default autotracking behaviour (eager).

For apps using v4 this argument will be ignored.

addon/components/basic-dropdown-content.ts
```diff
@@ -137,61 +137,65 @@ export default class BasicDropdownContent extends Component<Args> {
...
 
-  @action
-  animateIn(dropdownElement: Element): void {
-    if (!this.animationEnabled) return;
-    waitForAnimations(dropdownElement, () => {
-      this.animationClass = this.transitionedInClass;
+  initiallyReposition = modifier(() => {
+    // Escape autotracking frame and avoid backtracking re-render
+    Promise.resolve().then(() => {
+      this.args.dropdown.actions.reposition()
```
gilest, line 166: This is the only code change I made. Since the `did-insert` modifier does not participate in auto tracking this was not an issue before.

I'm reasonably confident that this change is compatible since the tests pass. If this modifier doesn't run a number of tests will fail.

Someone more familiar with this library may have a better suggestion to avoid doing this operation with a derived data pattern (since it's a side-effect type modfifier).

gilest:
@mkszepp would you mind reviewing this?

PR #716, opened 2023-11-13
Upgrade ember cli to 4.12.2

mkszepp:


mkszepp:
got the same bug like some months ago... `ember-cli-memory-leak-detector` is not compatible with `ember-qunit`.

While there is not fixed https://github.com/steveszc/ember-cli-memory-leak-detector/issues/49, we do remove this addon

PR #722, opened 2023-11-17
Update devDependencies

mkszepp:
* drop ember-code-example-component (replaced with own component)
* drop simple-git


mkszepp:
it was expected that `embroider-optimized` fails... (see https://github.com/ember-fastboot/ember-cli-fastboot/issues/925)

Next step should be to remove `ember-cli-fastboot` or make a patch like here https://github.com/NullVoxPopuli/ember-page-title/commit/bf8310a1cde66d0275b1891a8e0998ee8973ef28

PR #789, opened 2024-02-14
Add shadow DOM support

herzzanu:
As started and discussed in [this PR](https://github.com/cibernox/ember-basic-dropdown/pull/618), we're adding support for shadow DOM.


apellerano-pw:
@mkszepp could this PR land on top of v7.3.0 since v8 is still in beta? I don't think it will have merge conflicts with v7 since it's before the switch to v2 addon

apellerano-pw:
After doing a spike today, it's non-trivial to port the shadow-dom support to v8.

I was able to apply the patch to basic-dropdown-content.ts no problem.

But for basic-dropdown.ts it runs into the same situation currently being hacked around here:

https://github.com/cibernox/ember-basic-dropdown/blob/d4ee8ef2346bfbe7eadc9697de0fbaf60448e17c/ember-basic-dropdown/src/components/basic-dropdown.ts#L352-L354

I believe we'll need a repeatable, type safe way to access `rootElement` in the v2 addon paradigm, as all the other `document.querySelector` usages in the file need to alternatively do `rootElement.querySelector` for shadow-dom support.

Not sure if this is already planned as part of the v2 addon migration. If so, the shadow-dom patch for v8 should wait until that work completes.

(The big idea behind this feature is that sometimes the Ember App is inside a shadow-dom and `document.querySelector` can't pierce thru. but the `rootElement` will be inside the shadow-dom, so any selector queries that begin with it will still work.)

mkszepp:
Thanks for PR and detail reporting/analysing...

> @mkszepp could this PR land on top of v7.3.0 since v8 is still in beta? I don't think it will have merge conflicts with v7 since it's before the switch to v2 addon
> 
> would you be open to one more v7 release for this change?

The v8 is more/less ready... i have only looked in the last time to close open PR / issues, specially parts which were braking, because we have already braking change with migration to v2 addon and so we can add additional braking changes before we release.
As i have seen shadow dom i have think to wait a little bit with the next major release, because i'm not sure if this is braking. I don't like when peoples are using forks for long time or generally, when it is possible to bring this features also into the main addon.

> 
> I was able to apply the patch to basic-dropdown-content.ts no problem.
> 
> But for basic-dropdown.ts it runs into the same situation currently being hacked around here:
> 
> https://github.com/cibernox/ember-basic-dropdown/blob/d4ee8ef2346bfbe7eadc9697de0fbaf60448e17c/ember-basic-dropdown/src/components/basic-dropdown.ts#L352-L354
> 

This line are not really a hack... its a correct ways which are using also other addons... The only part problem in this case is, that the declarations are not perfect and so we need a ts-ignore line. In earlier versions there was used `ember-get-config` on this place to get TS valid instead of making a disabled rule, since it get fixed.


If you can prepare everything in v7 and adding also some tests which are passing, i will look to find time to bring this also in v8.

apellerano-pw:
@mkszepp can you create a new branch at the head of v7? This PR as written should be exactly what is needed, however the base branch needs to be modified to point to a v7 branch, since `master` now holds v8

mkszepp:
@herzzanu @apellerano-pw i have now created an new PR (#791) with all this changes, but based on v8 version (master branch). It was more/less impossible to bring this changes to v8 version.
The testes are looking good with this changes... now we should add some tests for shadow dom, before we can merge it

mkszepp:
i have tested a little bit and seen that `owner.rootElement.querySelector?` is in test-app generally null... is there missing any part in my app which you have added on basic dropdown call? see https://github.com/cibernox/ember-basic-dropdown/pull/791/files#diff-6d97234033770229316a952c6805250b59aec687f2d1dafa45edca0759e61630

apellerano-pw:
In our case we use an instance-initializer to wrap the Ember app's existing `rootElement` in a shadow DOM and then set that as the new `rootElement`.

For broader context, this is a browser extension which inserts an ember app into other pages. In that scenario you need to protect the entire Ember subtree from DOM modifications in the host application (e.g. if host calls `body.normalize()` it will destroy nodes Ember uses for bookkeeping.) So you hide it inside a shadow DOM.

This is why the equivalent to `document.querySelector()` in a normal application, would be `rootElement.querySelector()` in a shadow DOM protected Ember app.

Your test has a really nice component for doing shadow DOM isolation at the component level, but this won't alter the behavior of the Ember app's `rootElement` so it'll always be null for you.

mkszepp:
I have looked a little bit around and i think we can replace `owner.rootElement` with `node.getRootNode()`...
I'm not 100% sure but i think this should do the same and it should work in my test example...
I will try to replace it on my branch.

Maybe you can also test in you app, if you have the same result with this replacing?

https://developer.mozilla.org/en-US/docs/Web/API/Node/getRootNode

apellerano-pw:
I gave `node.getRootNode()` a shot, but I don't think it will be a silver bullet. You need an Element reference to be `node`, and those aren't readily available in Ember Octane.

One way is to use `did-insert` modifier in the template. However, the templates don't offer a good spot to do that. basic-dropdown's template can be element-less. basic-dropdown-content's template has two conditional root elements.

To continue down this path, I think basic-dropdown would need to add some element to the DOM specifically for the purpose of finding the root node. E.g.

```gjs
@tracked rootNodeFinderElem;

get rootNode() {
  return this.rootNodeFinderElem.getRootNode();
}

@action setRootNodeFinder(elem) {
  this.rootNodeFinderElem = elem;
}

<template>
  <div {{did-insert this.setRootNodeFinder}}></div>
  {{!--
    ...existing basic-dropdown template...
  --}}
</template>
```

It could then pass `rootNode` down to the trigger and content components so they don't have to do the lookup again themselves.

apellerano-pw:
Here's a POC of the `getRootNode()` trick on v7. Seems to work as well as `getOwner()` in my shadow dom use case

https://github.com/EmberExperts/ember-basic-dropdown/compare/shadow-dom...ProsperWorks:ember-basic-dropdown:get-root-node

mkszepp:
@apellerano-pw thank for your testing... passing rootNode is maybe one way, but maybe we find a better way.

An other idea/way from my side is, that we can extend the publicAPI with `registerTrigger` & if necessary a `registerContent` and `registerWormhole`.  The `registerTrigger` we can call directly in our `basic-dropdown-trigger´ modifier... for content maybe we can add an additional modifier which is handling everything

In this case we don't need to call any `.querySelector` for searching the element...

Registering the elements is maybe for v8 optional, so we are not braking for now... in future we can also discuss if every custom trigger/content needs to make register the components.

Its only a idea, not tested and i'm also not sure if we have every case covered... I will try to make a draft in next days / maybe also in later today, when i have some time.

I think this could be a good solution, because the current way is not working always (like there was in my case)

mkszepp:
Added `register` function on PR #791, see commit https://github.com/cibernox/ember-basic-dropdown/pull/791/commits/a5f7b3a8d5dabeb7fb850dd7181e9465f795fc18

mkszepp:
@herzzanu @apellerano-pw i have no prepared everything on PR #791 and added two tests (one with destination outside of shadow and one inside)

Feel free to review it.. the plan is to merge it in the next days if everything is ok

mkszepp:
feature landed with PR #791

PR #791, opened 2024-02-16
Add shadow DOM support

mkszepp:
The basic ideas were started in PR #618 & #789 which are based on v7 branch.

While porting to v8, we have found an other solution to make it possible to use the addon in shadow dom.

In generally everything is working out of the box, no addition configurations are necessary to use the addon in an shadow dom.
After this PR close the addon can be used partially or also completly in an shadow dom.

**Which changes were done?**
- Instead of using  `.querySelector` to find the elements, there was added a registerApi (`publicAPI.actions.registerTriggerElement` & `publicAPI.actions.registerDropdownElement`). This is atm optionally, so this is not braking for consumer apps. The addon is already the registerApis and the fallback to `.querySelector` are still present. For getting work with shadow dom there is required to use the registerApi.
- `args.destinationElement` was added... normally there is not necessary to pass the Element, but maybe in some cases it is necessary for example when the trigger is inside an other shadow root as the destination in which the content should be added.
- We have added in the tests some shadow dom simulations
- The test-app can be started with `ember s shadowDom` / `ember test shadowDom` to simulate the complete app in an shadow dom


apellerano-pw (commented):
This lgtm. My shadow dom use case is accurately mimicked when the `shadowDom` process arg is applied.

Will there be an ember-try scenario setting that flag? Or some other way to incorporate into CI?

I'm unfortunately not able to test these changes with my current repo because I can't yet upgrade to v8 (some build issue with my project, ember-auto-import, and this addon; not your problem). But when I do get to do the upgrade, if I find any bugs I'm confident I can patch it on top of these changes and submit a PR.

herzzanu (approved):

mkszepp:
Thank for reviewing!

`ember s shadowDom` is more/less similare like the usecase in your app.
My first plan was to use the complete test-app one time like now and one time everything in an shadow dom, but bringing the tests to render inside a shadow dom is a little bit tricky... i have tested some ways, but unfortunately they were not working... `this.element` was always outside the shadow dom or he doesn't find the shadow element id...
Maybe someone has a good idea which is working, so i will implment also a ember-try szenario which is running in shadow dom.


> But when I do get to do the upgrade, if I find any bugs I'm confident I can patch it on top of these changes and submit a PR.
That's exactly what I had in mind. I think we can merge it and when there a fixes necessary you can make a PR.

apellerano-pw:
In our case we use an instance-initializer to make the swap to shadow-dom, and this plays nicely in the test environment. Here's what `this.element` points to in a test context:

image

instance-initializer is structured like so

```js
initialize(appInstance) {
  let appRootElement = appInstance.get('rootElement');
  if (typeof appRootElement === 'string') {
    appRootElement = document.querySelector(appRootElement);
  }
  const targetElement = appRootElement || document.getElementsByTagName('body')[0];

  const hostElement = document.createElement('div');
  hostElement.attachShadow({ mode: 'open' });

  const rootElement = document.createElement('div');

  hostElement.shadowRoot.appendChild(rootElement);
  targetElement.appendChild(hostElement);

  // stuff to insert css into shadow dom...

  appInstance.set('rootElement', rootElement);
}

mkszepp:
thanks for the snippet... tested shortly and it looks that is working in test-app 🎉... we need only to update +/- all tests because we need to passt rootElement in every `assert.dom`... i will look to find time in next days

mkszepp:
@apellerano-pw @herzzanu `test-app` can now also run completly in shadow dom and the added `Tests Shadow dom` in ci is also passing all tests 🎉

So we are ready for landing this feature. If you find any bugs in your app after releasing, let me know and we will try to fix it and adding additional tests.
I will make a release in the next days.

Thanks for helping!

PR #903, opened 2024-05-30
Document @rootEventType

ArnaudWeyts:
### Summary

I noticed this argument wasn't documented.

### Description

This is related to https://github.com/cibernox/ember-power-select/pull/1819, where we now set this property.


mkszepp (approved):
Thank you, LGTM

PR #927, opened 2024-06-18
Fix dropdown not correctly positioned horizontally when body position is relative

ArnaudWeyts:
## Why

We noticed that in certain use cases the dropdown is not correctly positioned horizontally when the body is set to `position: relative` and you scroll horizontally. When digging through some issues and the code I discovered this [fix](https://github.com/cibernox/ember-basic-dropdown/pull/333) for vertical positioning. I believe that we should also implement a similar fix for horizontal positioning. We've patched this dependency in our own app and have verified that it works.

Reproducible use case (scroll to the right, then open a dropdown): https://stackblitz.com/edit/ember-cli-editor-output-1hdarw?file=app%2Ftemplates%2Fapplication.hbs,app%2Fstyles%2Fapp.css,app%2Fapp.js&title=Ember%20Starter

Screenshot 2024-06-18 at 15 37 38

### DIFF
--- a/ember-basic-dropdown/src/utils/calculate-position.ts
+++ b/ember-basic-dropdown/src/utils/calculate-position.ts
@@ -93,8 +93,21 @@ export const calculateWormholedPosition: CalculatePosition = (
     style.width = dropdownWidth;
   }
 
+  /**
+   * Fixes bug where the dropdown always stays on the same position on the screen when
+   * the <body> is relatively positioned
+   */
+  const isBodyPositionRelative =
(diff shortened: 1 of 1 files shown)


mkszepp (approved):
Thanks, LGTM

ArnaudWeyts:
Hi @mkszepp I see we didn't create a new release yet that includes this change, any chance we could?

mkszepp:
Sry i will publish everything on begin of next week

mkszepp:
@ArnaudWeyts now released in v8.2.0

Issue #931, opened 2024-07-25
Remove code that specifies different behaviour in tests inside of BasicDropdownWormhole


ArnaudWeyts:
# Why 

Our visual tests render a duplicate `#ember-testing` container because of `BasicDropdownWormhole`

# What

https://github.com/cibernox/ember-basic-dropdown/blob/39b4a0e103ce7ac44890646ed11803dc3f8ff6b7/ember-basic-dropdown/src/components/basic-dropdown-wormhole.ts#L22-L29

This part of the code will result in the following structure for visual tests:

```html
<div id="ember-testing">
  <div id="ember-testing"></div>
</div>
```

I believe that this should actually be removed (except for the fastboot stuff possibly?), since visual tests use the `application.hbs` template, the wormhole should get rendered using the regular flow.

What might be required though is providing a test setup helper that does render the wormhole for regular rendering tests that don't use the `application.hbs` template.

What we've done for now as a workaround is not rendering the `BasicDropdownWormhole` in `application.hbs` in a testing environment and instead setting up the required HTML ourselves in a test helper:

```ts
import { getRootElement } from '@ember/test-helpers';

export function setupWormhole(hooks: NestedHooks) {
  hooks.beforeEach(function () {
    const wormholeBasicDropdown = document.createElement('div');
    wormholeBasicDropdown.id = 'ember-basic-dropdown-wormhole';
    getRootElement().appendChild(wormholeBasicDropdown);
  });

  hooks.afterEach(function () {
    const wormholeBasicDropdown = document.getElementById('ember-basic-dropdown-wormhole');
    wormholeBasicDropdown?.remove();
  });
}
```

This can then be setup just for rendering tests:

```js
// app/tests/helpers/index.ts
import {
  setupRenderingTest as upstreamSetupRenderingTest,
} from 'ember-qunit';

function setupRenderingTest(hooks: NestedHooks, options?: SetupTestOptions) {
  upstreamSetupRenderingTest(hooks, options);
  setupWormhole(hooks);
}
```

Would love to hear your thoughts on this, and thanks for the work on this project!

mkszepp:
released in ´v8.3.0`

PR #933, opened 2024-08-07
Remove test-specific behaviour for BasicDropdownWormhole

rohitpaulk:
Fixes https://github.com/cibernox/ember-basic-dropdown/issues/931. 

I tried reading through the history of this to figure out why it was present, but nothing turned up. The commit that added the test-specific condition was https://github.com/cibernox/ember-basic-dropdown/commit/067ca88c93622a583dfaaf7bd0363d1a34ec7801. 

cc: @mkszepp in case you've got more context on why this was needed?

If tests are using `application.hbs`, this should work as-is. If tests aren't using `application.hbs` (component tests for example), the wormhole will need to be setup manually anyway + it'll still need to have the wormhole-specific ID, not `ember-testing`.


mkszepp:
@rohitpaulk the code was added in this commit PR https://github.com/cibernox/ember-basic-dropdown/pull/743 ... i have only synced the code parts, no specific condition.

Can you add any test that we can see that it works fine also after this changes?

rohitpaulk:
@mkszepp thanks for the context! I've added tests for this. 

The commit you linked to indeed only re-arranges stuff and doesn't alter logic. https://github.com/cibernox/ember-basic-dropdown/commit/067ca88c93622a583dfaaf7bd0363d1a34ec7801 seems to be where the `isTesting` condition was added. That's the one I wasn't sure about - do you remember if there was a specific reason to add the `isTesting` check?

rohitpaulk:
Looks like there a shadow dom test failure, will check shortly!

mkszepp:
@rohitpaulk i have checked why the test fails and its not directly caused with you changes... its because we are adding for shadow dom tests already with a initializer the element (for tests unfortunately there is no better solution)

replacing your test `dom('#custom-wormhole-destination')` with `dom('.ember-application #custom-wormhole-destination')` solves the issue and test is also correct

https://github.com/cibernox/ember-basic-dropdown/blob/d816d191044a89e930676566661d5c1e0e45f3d7/test-app/app/instance-initializers/shadow-root.js#L22-L32

Unfortunately I have no rights to work on your branch, so I had to make a fork

Thanks for fixing the issue

PR #941, opened 2024-11-11
Allow `@glimmer/component` v2

mkszepp:


mkszepp:
Blocked by https://github.com/emberjs/ember.js/issues/20790

Techn1x:
I'm surprised the tests are failing even though glimmer/component v1 was specified for them in the ember-try config

mkszepp:
Its a problem inside mono repo with pnpm and one lock file...
We were running in same issue with test helper in power select... only fix is to modify ci script... in eps it has passed tests

mkszepp:
We force now `@glimmer/component` v1 for ember <= 4.8 cause of reported bug https://github.com/emberjs/ember.js/issues/20790. When the bug is fixed in glimmer v2 we can safely remove in ember-try...

mkszepp:
> > Its a problem inside mono repo with pnpm and one lock file...
> 
> wonder if pnpm 9 fixes it (probably a long shot though)

no i don't think so... its still a open PR (since years 😅) i think NVP has a other setup published for github pipelines... but not sure which changes we need for them

mkszepp:
@Techn1x released in `v8.4.0`

PR #946, opened 2025-02-07
Test: Prefer `config.destination` and default `ember-basic-dropdown-wormhole`

### LINKED ISSUE #943: Content is rendered into wrong element in tests
In testing environment the content of dropdowns is rendered into `config['APP']?.rootElement` (which is `ENV.APP.rootElement = '#ember-testing';` by default) instead of the defined `destination` or the default `ember-basic-dropdown-wormhole`.

In our acceptance tests this leads to wrongly placed content in tests. Should `config['APP']?.rootElement` be considered a fallback if neither `destination` nor `ember-basic-dropdown-wormhole` are available?

dagroe:
… over root element as destination in tests

Close #943


mkszepp (approved):
Thanks!

PR #1002, opened 2025-09-05
Remove use of `ember-lifeline`

bertdeblock:
- `ember-lifeline` is unmaintained, and is already triggering deprecation warnings in newer Ember versions
- The recommendation to use it is outdated, see [this issue](https://github.com/ember-cli/eslint-plugin-ember/issues/2185)
- `ember-basic-dropdown` is widely used, so the less dependencies the better

### DIFF
--- a/docs/app/components/snippets/content-events-1.js
+++ b/docs/app/components/snippets/content-events-1.js
@@ -1,6 +1,6 @@
 import Component from '@glimmer/component';
 import { action } from '@ember/object';
-import { runTask, cancelTask } from 'ember-lifeline';
+import { task, timeout } from 'ember-concurrency';
 
 export default class extends Component {
   notifications = [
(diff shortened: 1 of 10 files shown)


bertdeblock:
`ember-lifeline` imports from `ember` => https://github.com/ember-lifeline/ember-lifeline/blob/master/packages/ember-lifeline/src/utils/get-timeout-or-test-fallback.ts#L1, which should trigger the barrel file deprecation.

bertdeblock:
More specifically, this one => https://github.com/ember-lifeline/ember-lifeline/blob/master/packages/ember-lifeline/src/utils/get-timeout-or-test-fallback.ts#L1

mkszepp:
@bertdeblock released in `v8.7.0`

PR #1081, opened 2026-03-05
add a test for ember 3.28

mansona:
I noticed when evaluating an Ember upgrade that ember-basic-dropdown doesn't say that it has support for ember@3.28, but I had a hunch that there was nothing in the codebase that actually prevented it from working.

Now that it's a v2 addon and on the new v2 addon blueprint it's quite easy to see that it does still have support for 3.28 🎉 I don't know if y'all might want to consider merging this to add back the official support, it can be quite useful to people who are upgrading really old apps to 3.28 and then migrating to Vite

Edit: I ran this PR's CI on my own fork to show that it was actually working: https://github.com/mansona/ember-basic-dropdown/actions/runs/22722030628/job/65886672854 so you will see CI go green once you approve running the CI job 👍

### DIFF
--- a/.try.mjs
+++ b/.try.mjs
@@ -26,6 +26,21 @@ const compatDeps = {
 export default {
   scenarios: [
     {
+      name: 'ember-lts-3.28',
+      npm: {
+        devDependencies: {
+          'ember-source': '~3.28.0',
+          '@glimmer/component': '^1.0.0',
+          ...compatDeps,
+          'ember-cli': '~4.12.0',
+        },
+      },
+      env: {
+        ENABLE_COMPAT_BUILD: true,
+      },
+      files: compatFiles,
+    },
+    {
       name: 'ember-lts-4.12',
       npm: {
         devDependencies: {


mkszepp:
@mansona v8.x (which is right now the stable released) supports offical ember 3.28+. The main branch (which will the next release), i have dropped < 4.12.

The `ember-truth-helpers` support with v5.0.0 only 4.8+ (and we have this dependency). https://github.com/jmurphyau/ember-truth-helpers?tab=readme-ov-file#compatibility

So i think there is not really correct to keep this support in next major, otherwise we need to widen range truth-helpers, test-helpers... (which i have dropped right now). But we need also to test against them.

Very old ember project with 3.28, can use vite or webpack without any issue and deprecation warnings with 8.x. They can move with 8.x until 6.12 (and maybe also to 7.x without any issue)

An other issue is, that with ember-power-select project we will support in future only 4.12+, because we have moved to native helper function, dropped promise proxy support and a lot of other things which are possible breaking for < 4.12.

The ember-basic-dropdown v8 will never be compatible with ember-power-select v9, because SCSS things are breaking, and people have used always the same version from this addons. I'm not sure if this will be confusing addon consumers, when we support here still 3.28+ and on power-select 4.12+

mkszepp:
I have shortly tested in one of my apps and i can confirm, that users can move also with ember-basic-dropdown / ember-power-select v8.11 (which are already a v2 addon) to ember v7.0 (there a no issues / deprecation warnings).

For this reason i will close this PR to readd 3.28 into next major of basic dropdown (8.x)

